PR #18, opened 2025-10-04
fix: Handle socket.gaierror for invalid hostnames

zZedix:
This PR adds socket.gaierror handling to prevent crashes when hostname resolution fails


ImMohammad20000:
formater have error

zZedix:
fixed

PR #24, opened 2025-10-05
Fix/sqlalchemy notin cleanup

zZedix:
Updated SQL cleanup queries to use SQLAlchemy’s supported notin_ operators, preventing AttributeErrors that previously stopped subscription/pruning jobs from deleting stale records


ImMohammad20000:
I`m not sure but i think its ok the way it wase

PR #67, opened 2025-10-31
feat: Add SS-2022 relay support

Max-Sum:
This is a prove of concept implementation.

SS-2022 has a feature called [Extensible Identity Headers](https://shadowsocks.org/doc/sip023.html). It allows relays without giving them the user PSKs. xray-core supports this feature and have [examples](https://github.com/XTLS/Xray-examples/blob/main/Shadowsocks-2022/README.ENG.md) on configuration file.

This PR adds support for relay only for SS-2022.

To use this feature, one have to add inbounds specifically for relay:
```
"inbounds": [
     {
       "tag": "SS2022-relay",
       "port": 1234,
       "protocol": "shadowsocks",
       "settings": {
         "method": "2022-blake3-aes-128-gcm",
         "password": "{{ relay psk }}",
         "clients": [
           {
             "address": "server",
             "port": 1234,
             "password": "{{ next hop server psk }}",
             "email": "my server"
           }
         ],
         "network": "tcp,udp"
       }
     }
   ],
```

In the host settings, there is a field to select relay inbounds.
The final password for user subscription will be `{{ relay psk 0 }}:{{ relay psk 1 }}:..:{{ server spk }}:{{ user psk }}`.

## Summary by CodeRabbit

* **New Features**
  * Added Shadowsocks 2022 relay inbounds selection and management in host configuration.
  * Enabled ordered relay password support for Shadowsocks 2022 subscription generation.
  * Added validation to ensure relay inbound chains are valid.

* **Bug Fixes**
  * Removed duplicate relay settings block in the host modal.

* **Documentation**
  * Added localization for relay settings (EN, FA, RU, ZH).

### DIFF
--- a/app/db/compiles_types.py
+++ b/app/db/compiles_types.py
@@ -49,7 +49,7 @@ class EnumArray(TypeDecorator):
 
 
 class StringArray(TypeDecorator):
-    """Custom SQLAlchemy type to handle String lists as a comma-separated string."""
+    """Custom SQLAlchemy type to handle String sets as a comma-separated string."""
 
     impl = String
 
@@ -65,11 +65,32 @@ class StringArray(TypeDecorator):
     def process_result_value(self, value, dialect):
         """Convert stored comma-separated string back to a StringArraySet."""
         if value is None:
-            return set()
+            return set[str]()
         if isinstance(value, str):
-            return set(v for v in value.split(",") if v)
-        return set(value)
+            return set[str](v for v in value.split(",") if v)
+        return set[str](value)
 
+class StringList(TypeDecorator):
+    """Custom SQLAlchemy type to handle ordered String lists as a comma-separated string."""
+
+    impl = String
+
+    def __init__(self, length=255, **kwargs):
+        super(StringList, self).__init__(length=length, **kwargs)
+
+    def process_bind_param(self, value, dialect):
+        """Convert list to a comma-separated string for storage."""
+        if not value:
+            return ""
+        return ",".join([str(v) for v in value])
+
+    def process_result_value(self, value, dialect):
+        """Convert stored comma-separated string back to a String list."""
+        if value is None:
+            return list[str]()
+        if isinstance(value, str):
+            return list[str](v for v in value.split(",") if v)   
+        return list[str](value)
 
 class DaysDiff(FunctionElement):
     type = Numeric()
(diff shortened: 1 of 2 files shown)


ImMohammad20000:
Do you test this?
Is it work currecly?

app/operation/host.py
```diff
@@ -107,3 +112,20 @@ async def modify_hosts(
         asyncio.create_task(notification.modify_hosts(admin.username))
 
         return await get_hosts(db=db)
+
+    async def _validate_ss2022_relay_chain(self, host: CreateHost) -> None:
+        tags = getattr(host, "ss2022_relay_inbound_tags", None)
```
M03ED: you have a pydantic model in here and you can just use the values directly, no need for getattr

M03ED:
we wait few days for changes, after that if there is no response i close this pr

Max-Sum:
> we wait few days for changes, after that if there is no response i close this pr

Will push changes this week. I was busy working on other projects.

app/db/models.py
```diff
@@ -448,6 +448,10 @@ class ProxyHost(Base):
         EnumArray(UserStatus, 60), default=list, server_default=""
     )
     ech_config_list: Mapped[Optional[str]] = mapped_column(String(512), default=None)
+    # Ordered relay inbound tags for Shadowsocks 2022
+    ss2022_relay_inbound_tags: Mapped[Optional[list[str]]] = mapped_column(
+        StringArray(256), default=None, nullable=True
```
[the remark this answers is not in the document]
M03ED: nope, StringArray type don't work this way

app/db/compiles_types.py
```diff
@@ -65,11 +65,32 @@ def process_bind_param(self, value, dialect):
...
         if value is None:
-            return set()
+            return set[str]()
         if isinstance(value, str):
-            return set(v for v in value.split(",") if v)
-        return set(value)
+            return set[str](v for v in value.split(",") if v)
+        return set[str](value)
 
+class StringList(TypeDecorator):
```
M03ED: why are you adding an extra model for no reason?

Max-Sum: StringArray is not order-preserving because it decodes to set instead of list.

Max-Sum: Relay servers are chained to provide a path to final server. The order reflects how the packets go through, therefore order is inherently important.
Inbounds that act as relay need to be excluded because the "client" field in them is filled with the information of next hop instead of the panel managed clients. User will need to fill them manually.

ImMohammad20000:
this pr is not compleated and open for a long time i have to close it

PR #104, opened 2025-11-25
fix: adjusted singbox template to the latest format

iamtheted:
## Summary by CodeRabbit

* **Enhancements**
  * Updated DNS configuration with simplified per-server setup.
  * Added Best Latency routing option for optimized connection selection.
  * Refined routing configuration for improved network handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/app/templates/singbox/default.json
+++ b/app/templates/singbox/default.json
@@ -14,8 +14,7 @@
         "tag": "dns-local",
         "type": "local"
       }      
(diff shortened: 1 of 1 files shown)


ImMohammad20000:
The ios version have problems with apple and i plan to change template system in #286 so we no longer need this

PR #106, opened 2025-11-25
Add Permission Manager to core structure

zZedix:
Added a basic permission manager with support for traffic limit,
user count limit and max expiry rules


ImMohammad20000:
this is not what i want

PR #298, opened 2026-02-24
fix(dashboard): fix config copy failing silently in Safari (iOS & macOS)

MatinDehghanian:
## Problem

Copying subscription configs (links, xray, outline, etc.) from the user action buttons appeared to succeed (toast showed "Copied") but nothing was actually written to the clipboard on Safari — both iOS and macOS.

**Root cause:** Safari requires `clipboard.write()` / `clipboard.writeText()` to be called synchronously within a user gesture handler. The previous implementation awaited `fetchContent(link)` before calling `copy()`, which yielded the JS event loop and broke Safari's user-gesture requirement, causing the clipboard write to silently fail.

## Fix

Use `navigator.clipboard.write()` with a `ClipboardItem` that wraps the fetch in a `Promise`. This initiates the clipboard write synchronously (satisfying Safari's security check) while the content resolves asynchronously. Falls back to the previous `writeText` approach for browsers without `ClipboardItem` support (e.g. older Firefox).

## Affected file

- `dashboard/src/components/users/action-buttons.tsx` — `handleLinksCopy`


ImMohammad20000:
reset to first commit and open another pr for fix(dashboard): use ClipboardItem to fix config copy in Safari commit

PR #335, opened 2026-03-29
feat: add xtls-rprx-vision-udp443 flow support

KlimSemikin:
Introduced support for `xtls-rprx-vision-udp443` flow across backend and frontend

Since this flow is client-specific and not supported by Xray inbound configuration, it is replaced with `xtls-rprx-vision` when generating node configs

## Release Notes

* **New Features**
  * Added 'xtls-rprx-vision-udp443' as a new VLESS flow option for proxy configuration. Users can now select this new flow type when creating or modifying proxies and user templates. The option is fully integrated across all proxy configuration interfaces and provides seamless compatibility with existing proxy settings.


ImMohammad20000:
Change target branch to dev

PR #347, opened 2026-04-01
feat(subscription): add xhttp support for Clash Meta subscriptions

Nemu-x:
## Summary
Adds xhttp transport support for Clash Meta subscription output.

## Motivation
Recent Mihomo (Clash Meta) releases introduced support for xhttp transport, which is already used in modern VLESS deployments.

This PR enables PasarGuard to generate compatible configurations for clients supporting xhttp.

## Changes
- add `XHTTPTransportConfig`
- register `xhttp` transport handler
- implement `_transport_xhttp`
- allow xhttp for `ClashMetaConfiguration`
- keep xhttp unsupported for regular Clash

## References
- Mihomo releases: https://github.com/MetaCubeX/mihomo/releases

### DIFF
--- a/app/subscription/clash.py
+++ b/app/subscription/clash.py
@@ -47,6 +47,7 @@ class ClashConfiguration(BaseSubscription):
             "tcp": self._transport_tcp,
             "raw": self._transport_tcp,
             "xhttp": self._transport_xhttp,
+            "splithttp": self._transport_xhttp,
         }
 
         # Registry for protocol builders
@@ -194,6 +195,10 @@ class ClashConfiguration(BaseSubscription):
     ):
         """Apply transport settings using registry"""
         network = inbound.network
+        
+        # Normalize legacy splithttp -> xhttp
+        if network == "splithttp":
+            network = "xhttp"
 
         # Normalize network type for clash
         if network in ("http", "h2", "h3"):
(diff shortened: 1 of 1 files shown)


M03ED:
Splithttp is old name of xhttp, you have to remove that too

M03ED:
And I think you should parse download settings into a clash acceptable format

Nemu-x:
> And I think you should parse download settings into a clash acceptable format

Thanks. I already fixed `splithttp` as a legacy alias of `xhttp`.

For `download-settings`, I tested the generated config with Mihomo 1.19.22 (clash verge client) and it is accepted and works correctly. From Mihomo's xhttp config structure, `download-settings` is a nested xhttp config object, so the current shape is intentionally passed through as nested xhttp settings.

If you want, I can still add explicit normalization for the nested `download-settings` fields to make the output stricter and more self-documenting.

ImMohammad20000:
Can you send the document of xhttp for clash meta

Nemu-x:
There is no dedicated public documentation page for `xhttp` in Mihomo yet.

The references I used are based on the actual Mihomo implementation and related PRs where `xhttp`  support was introduced:

* https://github.com/MetaCubeX/mihomo/pull/2655
* https://github.com/MetaCubeX/mihomo/pull/2647
* https://github.com/MetaCubeX/mihomo/pull/2645

From these changes, `xhttp` is implemented as a transport with fields like `mode`, `headers`, `no-grpc-header`, `x-padding-bytes`, and `download-settings`, where `download-settings` itself is a nested xhttp config.

So the current output shape follows the actual Mihomo config structure.

I also tested the generated config with Mihomo 1.19.22 (Clash Verge), and it is accepted and works correctly.

If needed, I can still add explicit normalization for `download-settings` to make the output more strict and self-documenting.

Issue #287, opened 2026-02-22
Migration Issue: Expiration Time Becomes Unlimited After Transfer from x-ui (SQLite3)


monhacer:
Hello,
I am trying to migrate my users from x-ui to Pasarguard, including their remaining traffic quota, remaining time (expiration date), and subscription data.
I have previously performed this migration, but after the transfer all users’ expiration dates became unlimited, and only the traffic quotas were migrated correctly.
I am using the SQLite3 version.
Could you please guide me on how to properly migrate the expiration time along with the traffic data?
If this is a bug in the migration script, I would appreciate it if you could investigate and fix it.
Thank you in advance.

PR #366, opened 2026-04-10
fix: optimistic cache updates for user CRUD to prevent scroll reset

logicalangel:
## Problem

On the users page, when a user is deleted, created, or modified, the user list fully refreshes. If you're editing multiple users mid-page, the list scrolls back to the top after each operation, forcing you to scroll down again to find the next user. This makes the UX very frustrating.

**Related issue:** #287 (UX improvements)

## Root Cause

- `useGetUsers` was configured with `staleTime: 0, gcTime: 0` — zero caching means every `invalidateQueries` triggers an immediate refetch that discards the current data
- Delete operations called `invalidateQueries({ queryKey: ['/api/users'] })` causing a full list refetch
- Create operations also did full invalidation instead of updating the cache
- The page-level `refreshAllUserData` used mismatched query keys (`'getUsers'`, `'/api/users/'`) that didn't match the actual key (`'/api/users'`)

### Cache configuration
- Set `staleTime: 10_000` (10s) and `gcTime: 5 * 60_000` (5min) on `useGetUsers` so data stays visible during background refetches

### Optimistic delete (`action-buttons.tsx`)
- Instead of `invalidateQueries`, now removes the user from cached data via `setQueriesData` and decrements `total`
- Stats/usage queries still invalidated since counts changed

### Optimistic create (`user-modal.tsx`)
- Instead of `invalidateQueries`, now prepends the new user to cached list and increments `total`

### Page-level refresh fix (`_dashboard.users.tsx`)
- Fixed `refreshAllUserData` to only invalidate stats queries (the old query keys were mismatched anyway)

### Error boundary
- Added reusable `ErrorBoundary` component wrapping `UsersTable` for graceful error recovery with retry button

## Testing

- TypeScript type-check: `npx tsc --noEmit` — zero errors
- Vite build: passes successfully
- Manual testing: edit/delete/create operations no longer cause scroll-to-top

## Summary by CodeRabbit

* **New Features**
  * Added error handling for the users table with a fallback message and retry option when loading fails.

* **Bug Fixes**
  * Improved data synchronization after creating or deleting users with optimistic cache updates.

* **Performance Improvements**
  * Optimized data caching and refresh timing for the users page.

* **Chores**
  * Bumped version to 2.2.1.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pasarguard-dashboard",
   "private": true,
-  "version": "2.2.0",
+  "version": "2.2.1",
   "scripts": {
     "dev": "vite dev",
     "build": "vite build",
--- a/dashboard/src/components/common/error-boundary.tsx
+++ b/dashboard/src/components/common/error-boundary.tsx
@@ -1,0 +1,55 @@
+import { Alert, AlertDescription, AlertTitle } from '@/components/ui/alert'
+import { Button } from '@/components/ui/button'
+import { AlertCircle, RefreshCcw } from 'lucide-react'
+import { Component, type ErrorInfo, type ReactNode } from 'react'
+
+interface Props {
+  children: ReactNode
+  fallbackTitle?: string
+  fallbackDescription?: string
+}
+
+interface State {
+  hasError: boolean
+}
+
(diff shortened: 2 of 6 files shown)


M03ED:
Change branch to dev

x0sina:
This problem already fixed on dev branch

PR #405, opened 2026-04-23
feat(client-template): add possibility to use vless flow and uuid in client template

KlimSemikin:
Add possibility to use uuid and flow variables in client template. You can add second outbound with same user uuid with this changes. It can be used with observer for example and pick different oubounds by ping etc
```json
  "outbounds": [
    {
      "protocol": "vless",
      "settings": {
        "vnext": [
          {
            "address": "...",
            "port": 443,
            "users": [
              {
                "encryption": "none",
                "flow": "{{ flow }}",
                "id": "{{ uuid }}"
              }
            ]
          }
        ]
      },
      ...
      "tag": "germany"
    }
```

### DIFF
--- a/app/subscription/xray.py
+++ b/app/subscription/xray.py
@@ -56,8 +56,8 @@ class XrayConfiguration(BaseSubscription):
             "wireguard": self._build_wireguard,
         }
 
-    def add_config(self, remarks, outbounds, template_content: str | None = None):
-        rendered_template = render_template_string(template_content) if template_content is not None else self.template
+    def add_config(self, remarks, outbounds, settings: dict, template_content: str | None = None):
+        rendered_template = render_template_string(template_content, context={"uuid": settings["id"], "flow": settings["flow"]}) if template_content is not None else self.template
         json_template = json.loads(rendered_template)
         json_template["remarks"] = remarks
         json_template["outbounds"] = outbounds + json_template["outbounds"]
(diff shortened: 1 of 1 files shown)


M03ED:
And why you should do this ?

KlimSemikin:
> And why you should do this ?

I use it like this one:
```json
{
  "inbounds": [
    {
      ... 
    }
  ],
  "outbounds": [
    {
      // main config from panel
      "tag": "proxy"
    },
    {
      "protocol": "vless",
      "settings": {
        "vnext": [
          {
            "address": "domain2.xyz",
            "port": 443,
            "users": [
              {
                "encryption": "none",
                "flow": "{{ flow }}",
                "id": "{{ uuid }}"
              }
            ]
          }
        ]
      },
      "streamSettings": {
        "network": "raw",
        "security": "tls",
          "fingerprint": "chrome",
          "serverName": "domain2.xyz"
        }
      },
      "tag": "germany"
    },
    {
      "tag": "direct",
      "protocol": "freedom"
    },
    {
      "tag": "block",
      "protocol": "blackhole"
    }
  ],
  "observatory": {
    "subjectSelector": [
      "germany"
    ],
    "probeUrl": "https://captive.apple.com/hotspot-detect.html",
    "probeInterval": "10s",
    "enableConcurrency": false
  },
  "routing": {
    "domainStrategy": "IPIfNonMatch",
    "balancers": [
      {
        "tag": "balancer",
        "selector": [
          "germany"
        ],
        "fallbackTag": "proxy"
      }
    ],
    "rules": [
      {
        "inboundTag": [
          "socks"
        ],
        "balancerTag": "balancer"
      }
    ]
  }
}
```
Main outbound "proxy" that added automatically by pasarguard uses CDN. I want to use cdn with paid traffic only for mobile networks with enabled whitelist blocks. So i don`t want to pay for users forgot enable another default config if they using wi-fi (without whitelist blocks). "proxy" will be used only if "germany" is not available. With {{ uuid }} second config will be related to same user with all trafic counting and so on

With this feature we get more possibilities for customise client configs and create more flexible configs with many outbounds related with current user

PR #416, opened 2026-04-26
fix(jobs): online user count under node timeouts

x0sina:
## Summary
- Prevent usage recording jobs from overlapping by enabling `coalesce` and `max_instances=1`
- Separate node API concurrency from DB write concurrency so slow nodes do not block healthy nodes
- Make the online users window configurable
- Document new usage recording knobs in `.env.example`

## Why
When a few nodes time out rapidly, they can delay usage collection long enough for all `users.online_at` values to fall outside the old 2-minute dashboard window. This makes total online users temporarily show `0` even while users are connected.

## Summary by CodeRabbit

* **New Features**
  * Added configurable parameters for system performance tuning, including usage recording concurrency limits and online users tracking time window.

* **Chores**
  * Enhanced job scheduler with improved execution controls to prevent overlapping runs and improve reliability.


x0sina:
Verified and fixed.

`NODE_API_SEM` was previously held across the full stats-processing path: the node API request, thread-pool lookup, executor parsing, and invalid UID logging. That made the semaphore limit more than just active node network calls.

I narrowed the semaphore scope in both `get_users_stats` and `get_outbounds_stats` so it only wraps `await node.get_stats(...)`. Parsing and logging now happen after the node API slot is released.

This keeps `JOB_RECORD_USAGE_NODE_CONCURRENCY` focused on limiting concurrent node requests, while avoiding unnecessary blocking when responses are large or parsing takes time.

PR #458, opened 2026-05-10
fix(users): preserve scheduled traffic cycles across manual resets

Rerowros:
Manual traffic resets currently move the timestamp used by the periodic reset scheduler. This change records the reset source and keeps the scheduled traffic cycle independent of manual resets. It is based directly on current `dev` and does not require #756, #758, or #759.

## Behavior change

- Single-user, selected-user, and reset-all manual actions clear current usage without postponing the scheduled cycle.
- Scheduled resets and next-plan activation establish a new cycle boundary. The existing day/week/month/year intervals remain unchanged.
- Reset-all preserves usage-reset history and lifetime traffic instead of deleting that history. Existing status handling, optional chart cleanup, and next-plan removal behavior are preserved.
- Existing `next_plan` + `data_limit_reset_strategy` configurations remain accepted. Next-plan activation retains the existing direct/template plan policy; this PR does not introduce a prohibition on that combination.

This intentionally changes manual-reset behavior. Please keep the PR in draft while agreeing on the cycle/next-plan policy.

## API and dashboard

User and subscription metadata expose `last_traffic_reset_at`, `last_cycle_traffic_reset_at`, and `next_traffic_reset_at`. The user dialog shows these dates with English, Persian, Russian, and Chinese labels; the subscription page shows the next reset. Optimized user API reads calculate the dates alongside lifetime usage without loading reset history. The API client was regenerated from this standalone schema.

There are no unrelated dashboard date-parser or cleanup changes.

## Migration and concurrency

Migration `b8e4e47b9f2c` follows the current dev head `8e2f1a9c4b70`. Existing history and writes from older binaries receive `legacy`; historical rows are never guessed to be scheduled resets. The latest legacy reset anchors the cycle only until a scheduled/next-plan boundary exists. Downgrade preserves usage amounts and timestamps but removes source provenance; re-upgrade consequently marks those rows legacy.

Run the migration before starting the new application version, and update scheduler workers together to apply the new policy consistently. Other pending migration heads should be reconciled through the normal Alembic merge process when combined.

Reset paths lock users in stable order, read current user/next-plan state, and recheck scheduled eligibility from current locked history rows. This avoids stale MySQL `REPEATABLE READ` snapshots counting traffic twice or applying another scheduled reset. The `DateDiff` SQL expression now includes its operands in the SQLAlchemy cache key, preventing reuse of an earlier bound cycle date.

After a committed reset, history refresh replaces cached log values with their persisted timestamps. MySQL/MariaDB may round timestamps to seconds; retaining the old ORM value made metadata disagree with the stored history. The regression reproduces that identity-map problem deterministically on SQLite as well and fails without the production fix.

## Validation

- Standalone SQLite full suite: **585 passed, 5 skipped**, with the single existing FinalMask fixture failure reproduced on a clean export of dev `59dccc8d`.
- Separate integration export with only the independent #889 fixture correction (`690a47ea`): **586 passed, 5 skipped**. That correction is not part of this branch.
- Fresh Alembic upgrade, schema drift check, legacy-writer migration test, and downgrade/re-upgrade history preservation passed.
- Dashboard production build, all four locale JSON/UTF-8 checks, API generation idempotence, and Ruff on changed Python files passed. Full Ruff retains the existing `app/core/hosts.py` import-order finding from dev.
- Three deterministic MySQL/MariaDB `REPEATABLE READ` regressions cover a scheduler waiting for manual/scheduled resets and stale next-plan activation. On final head `b9360a91`, both database CI jobs pass the dedicated step containing these three regressions and the persisted-date regression before the full suite. The existing MySQL timestamp precision is preserved without depending on #758.
- Final MySQL/MariaDB CI: each dedicated step reports 4 passed; each full suite reports 588 passed, 2 skipped and only the unrelated FinalMask fixture failure fixed in #889. Final SQLite, PostgreSQL and TimescaleDB CI each report 585 passed, 5 skipped and the same baseline failure. All five jobs have completed.

### DIFF
--- a/.github/workflows/test-database-migrations.yml
+++ b/.github/workflows/test-database-migrations.yml
@@ -195,7 +195,7 @@ jobs:
               run: make check-migrations
 
             - name: Verify traffic reset races under REPEATABLE READ
-              run: .venv/bin/python -m pytest tests/api/test_user_usage_reset_cycle.py -k mysql_repeatable_read -q
+              run: .venv/bin/python -m pytest tests/api/test_user_usage_reset_cycle.py -k "mysql_repeatable_read or optimized_reset_dates" -q
 
             - name: Run Tests
               run: make test
@@ -321,7 +321,7 @@ jobs:
               run: make check-migrations
 
             - name: Verify traffic reset races under REPEATABLE READ
-              run: .venv/bin/python -m pytest tests/api/test_user_usage_reset_cycle.py -k mysql_repeatable_read -q
+              run: .venv/bin/python -m pytest tests/api/test_user_usage_reset_cycle.py -k "mysql_repeatable_read or optimized_reset_dates" -q
 
             - name: Run Tests (multi-worker all-in-one)
               env:
--- a/.github/workflows/test-database-migrations.yml
+++ b/.github/workflows/test-database-migrations.yml
@@ -194,6 +194,9 @@ jobs:
             - name: Check Migrations
               run: make check-migrations
 
+            - name: Verify traffic reset races under REPEATABLE READ
+              run: .venv/bin/python -m pytest tests/api/test_user_usage_reset_cycle.py -k mysql_repeatable_read -q
+
             - name: Run Tests
               run: make test
 
@@ -316,6 +319,9 @@ jobs:
 
             - name: Check Migrations
               run: make check-migrations
+
+            - name: Verify traffic reset races under REPEATABLE READ
+              run: .venv/bin/python -m pytest tests/api/test_user_usage_reset_cycle.py -k mysql_repeatable_read -q
 
             - name: Run Tests (multi-worker all-in-one)
               env:
--- a/app/app_factory.py
+++ b/app/app_factory.py
@@ -12,6 +12,7 @@ from app.middlewares import setup_middleware
 from app.nats import is_multi_worker, require_nats_if_multiworker
 from app.nats.message import MessageTopic
 from app.nats.router import router
+from app.node.errors import NodeRevocationError
 from app.settings import handle_settings_message
 from app.subscription.client_templates import handle_client_template_message
 from app.utils.logger import get_logger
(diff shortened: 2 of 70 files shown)


M03ED:
thanks for pr, there are some points i should write here 
first of all why these 3 pr are related ? they have completely different subjects
second is this behavior was implemented since day one and it will be counted as major change, 
third this remember me something i forgot to do, using next_plan and reset strategy at the same time should be rejected by api, its better for user to use only one of them (at some periods i had plan to remove next plan but that is not happening, at least today) and i don't know why people use 2 strategy at the same time

i check 2 other pr later but this pr stays open for a now, i will discuss it with other developers

Rerowros:
> thanks for pr, there are some points i should write here first of all why these 3 pr are related ? they have completely different subjects second is this behavior was implemented since day one and it will be counted as major change, third this remember me something i forgot to do, using next_plan and reset strategy at the same time should be rejected by api, its better for user to use only one of them (at some periods i had plan to remove next plan but that is not happening, at least today) and i don't know why people use 2 strategy at the same time

Thanks for reviewing.

About the behavior change: I agree this is user-visible and can be considered major. I proposed it because the current implementation uses the latest usage reset log as the cycle anchor, so a manual/API reset also postpones the user's monthly/weekly/daily/yearly reset.

Manual/API reset and scheduled reset have different meanings: manual/API reset is a one-off administrative action, while scheduled reset is the user's reset-cycle boundary. Right now both actions use the same anchor, so a manual reset can unexpectedly change the next scheduled reset.

I do not want to force this behavior if the project considers the old logic intentional. I can adjust the PR to make the behavior explicit/backward-compatible, for example by adding an option such as `reset_cycle` / `preserve_reset_cycle`, or by separating "reset usage only" from "reset usage and restart cycle" in the API/UI.

Regarding `next_plan + reset strategy`: I agree this combination is confusing. My intent with `reset_source=next_plan` was only to say that applying a next plan starts a new usage cycle. If the preferred direction is to reject `next_plan` together with a reset strategy, I can move that validation into a separate PR and keep this PR focused only on manual resets not silently changing the scheduled cycle.

I can also remove the unrelated dashboard date parsing changes from this PR to keep the scope smaller.

PR #466, opened 2026-05-11
feat(core-config): add unsaved changes confirmation dialog

ArthurKoba:
- Implement change tracking for core configuration editor
- Block modal close when unsaved changes exist
- Show confirmation dialog with options to keep editing or discard changes
- Add translations for all supported languages (en, ru, fa, zh)
- Reset dirty state after successful save operation

Solving the issue: https://github.com/PasarGuard/panel/issues/459

### DIFF
--- a/dashboard/public/statics/locales/en.json
+++ b/dashboard/public/statics/locales/en.json
@@ -1877,7 +1877,11 @@
     "shadowsocksPasswordGenerationFailed": "Failed to generate password",
     "realityTab": "Reality",
     "vlessTab": "VLESS",
-    "shadowsocksTab": "ShadowSocks"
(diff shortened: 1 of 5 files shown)


ImMohammad20000:
this will change in feauture

ArthurKoba:
Yes, I saw your message, but by the time I made the implementation. How soon are you planning to make changes to this section and release it? If it's not that soon or the process drags on, and there's an interim release or a fix in the main, then I think PR can be useful for at least some time.

ImMohammad20000:
There is no but in feature versions

ArthurKoba:
I hope there won't be a similar problem when editing kernels in the future.

PR #520, opened 2026-05-23
refactor: update admin and user endpoints to use IDs instead of usernames

ImMohammad20000:
following deprecated APIs in #457 default api format changed in version 5.0.0


M03ED:
we do it in version 6

ImMohammad20000:
Let it be open

PR #524, opened 2026-05-24
fix(decimal-input): Preserve zero value display

subzero911:
Bug: if you enter '0' (unlimited)
image

it disappears when user reopens the dialog:
image

This fix should handle it


x0sina:
You edited main shared component
I Fixed it in dev

PR #601, opened 2026-06-13
security+perf: subscription status validation and config caching

wahh3b-lgtm:
## Changes

- **Status validation** (\pp/operation/subscription.py\): Added \UserStatus.disabled/expired/limited\ checks on \/info\ and \/usage\ endpoints to prevent data leakage for inactive accounts
- **Subscription caching**: TTL-based in-memory cache (60s TTL, 512 max entries) for subscription config generation, avoiding redundant JSON parsing and base64 encoding on repeated requests
- **Cache key**: Includes user_id, status, client_type, quantized traffic (10MB buckets), and revoked timestamp for accurate invalidation

## Files Changed
- \pp/operation/subscription.py\ — status checks + subscription cache

## Verification
- [x] Ruff lint + format pass
- [x] 13/13 tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed security vulnerability allowing disabled, expired, or limited accounts to access subscription data.

* **Performance**
  * Optimized subscription request handling with intelligent caching to reduce latency and improve response times.

### DIFF
--- a/app/operation/subscription.py
+++ b/app/operation/subscription.py
@@ -1,3 +1,4 @@
+import time
 import re
 from json import dumps as json_dumps
 from typing import Any
@@ -12,7 +13,7 @@ from app.db.crud.hwid import (
     register_user_hwid,
 )
 from app.db.crud.user import get_user_usages, user_sub_update
-from app.db.models import User
+from app.db.models import User, UserStatus
 from app.models.admin import AdminDetails
 from app.models.settings import Application, ConfigFormat, HWIDSettings, SubRule, Subscription as SubSettings
 from app.models.stats import UserUsageStatsList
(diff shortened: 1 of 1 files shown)


M03ED:
i close this one since cache is almost useless in this scenario, data have to be in real time ( 10 mb chunks can become invalidate in seconds ) doing some python loop are not the hot path and just add unnecessary complexity 
also there is no implementation for nats and if we use multi worker each worker have its own cache and this can cause high memory usage

wahh3b-lgtm:
> i close this one since cache is almost useless in this scenario, data have to be in real time ( 10 mb chunks can become invalidate in seconds ) doing some python loop are not the hot path and just add unnecessary complexity also there is no implementation for nats and if we use multi worker each worker have its own cache and this can cause high memory usage

oke 👍

PR #605, opened 2026-06-14
perf: optimize database pools, batch commits, and node health checks

wahh3b-lgtm:
## Changes

- **DB connection pool** (`config.py`, `app/db/base.py`): Increased `pool_recycle` to 1800s, added configurable `pool_timeout` (default 15s) with `ge=1` validation, enabled `pool_use_lifo=True`
- **Performance indexes** (Alembic migration): Added composite indexes via proper migration (`a1b2c3d4e5f6`) — `(status, expire)`, `(status, used_traffic)`, `(node_id, created_at)` and others with `if_not_exists=True`
- **Record usages** (`app/jobs/record_usages.py`): Cached dialect lookup, batched node stats in single transaction with per-row `safe_execute` fallback that continues on individual failures
- **Node health checks** (`app/node/__init__.py`): Concurrent `asyncio.gather()` with snapshot revalidation and exception logging for failed probes
- **XRayConfig** (`app/core/xray.py`): Reverted TLS reads to synchronous (`run_until_complete` on a running loop always raised `RuntimeError`)
- **.env.example**: Updated pool defaults to match new config

## Files Changed

- `config.py`
- `app/db/base.py`
- `app/db/__init__.py`
- `app/db/migrations/versions/a1b2c3d4e5f6_add_performance_indexes.py` (new)
- `app/jobs/record_usages.py`
- `app/node/__init__.py`
- `app/core/xray.py`
- `.env.example`

## Verification

- [x] Ruff lint + format pass
- [x] 13/13 tests pass
- [x] No database schema changes beyond additive indexes

## Summary by CodeRabbit

* **New Features**
  * Added database performance indexes across multiple tables.
  * Added a configurable database connection pool timeout setting (with minimum validation).
* **Chores**
  * Updated the default database connection pool recycle setting from 300 to 1800 seconds.
  * Improved database connection pool behavior (including LIFO usage) to better manage connections.
* **Bug Fixes**
  * Improved node health checks and usage recording reliability during transient database issues.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -47,7 +47,8 @@ UVICORN_PORT = 8000
 ## Database pool settings are per worker, not global. Four workers with pool size 10 can open up to 40 base connections.
 # SQLALCHEMY_POOL_SIZE = 25
 # SQLALCHEMY_MAX_OVERFLOW = 60
-# SQLALCHEMY_POOL_RECYCLE = 300
+# SQLALCHEMY_POOL_RECYCLE = 1800
+# SQLALCHEMY_POOL_TIMEOUT = 15
 # ECHO_SQL_QUERIES = False
 
 ## NATS connection and subjects for multi-process/node coordination.
--- a/app/core/xray.py
+++ b/app/core/xray.py
@@ -372,7 +372,7 @@ class XRayConfig(dict):
         """Normalize Hysteria Salamander masks into finalmask for client generation."""
         finalmask = stream.get("finalmask") or stream.get("finalMask")
         if isinstance(finalmask, dict):
-            finalmask = deepcopy(finalmask)
+            finalmask = {k: v for k, v in finalmask.items()}
         else:
             finalmask = {}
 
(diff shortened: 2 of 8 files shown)


M03ED:
`_dialect_cache` race on first call:
Two coroutines can both see _dialect_cache is None, both enter GetDB(), both set the cache. Not a bug in practice (they'd both set it to the same value), but it can open two connections on startup instead of one.
using `engine.begin()` can only be done inside `safe_execute` that handle all errors, i see all of your pull requests targeg main branch, were not gonna accept any pr that target main branch

M03ED:
`idx_users_status` and `idx_users_status_expire` overlap — if you query `WHERE status = ? AND expire < ?`, only the composite one is needed and the single-column one becomes redundant overhead on writes. 
i don't know why temp_keys need index.
database migrations must be generated with help of `alembic` and indexes should be defined in db models so alembic can detect the changes

wahh3b-lgtm:
Superseded by #615 targeting dev branch

PR #662, opened 2026-06-28
feat(dashboard): select optional Xray API services in both core editors

Multi-Engineer-dev:
## What

Adds first-class support for selecting optional Xray gRPC **API services** from
the dashboard, in both core-config editors, plus the supporting validation,
persistence, and localization.

## Why

Xray exposes optional API services (e.g. `RoutingService`, `ObservatoryService`)
beyond the always-on `HandlerService` / `LoggerService` / `StatsService`.
Operators previously had to hand-edit raw JSON to enable them. This adds a
guided, validated control and ensures the selection round-trips correctly
through the structured editor.

## What's included

- **Allowlist + helpers** (`lib/xray-api-services.ts`): canonical required and
  optional service lists mirrored from xray-core, with pure helpers to read and
  toggle the selected optional services. Unit tests pin the allowlist so it
  can't drift silently.
- **Legacy core config modal**: an "API Services" control to toggle optional
  services, with blocking validation that rejects unknown names and de-dupes
  them, plus accessible label/checkbox associations.
- **Structured core editor**: a new "API" section mirroring the same control for
  the default route-based editor.
- **Localization**: the structured editor's API section header and card
  (title + hint) are translated across all four locales (en/fa/ru/zh); the
  "API" acronym is kept as-is.
- **Fixes**:
  - Toggling a service off now reliably enables Save. The structured editor's
    adapter re-emits `api` from `raw.source`, so toggles now update `raw.source`
    and `raw.topLevel` in lockstep (`setRawOptionalService`) — otherwise removing
    the last optional service produced no diff and left Save disabled.
  - The API section header renders proper labels even when a stale PWA-cached
    locale is served, via `defaultValue` fallbacks on `PageHeader`.

## Tests

- Unit tests for the allowlist helpers, including regressions for the
  toggle-persistence fix.
- All four locale files validated (parse + key resolution).
- `esbuild` integration check against a real core config, lints, and a
  production build all pass.

## Related
Dashboard half of the Xray API services feature, which spans three repos:
- **Node** — merges these user-selected services with the node's required ones
  (instead of overwriting), fixes the `ObservatoryService` startup crash, and adds
  the `RoutingService` RPCs: [Node PR 60](https://github.com/PasarGuard/node/pull/60).
- **Bridge** — Python async wrappers for the routing RPCs over gRPC + REST:
  [Bridge PR 16](https://github.com/PasarGuard/node_bridge_py/pull/16).
This PR ships the enable/disable toggles; the dashboard UI to *use* `RoutingService`
(manage rules, balancers, test routes) is a follow-up on top of the bridge.

## Summary by CodeRabbit

* **New Features**
  * Added an **API** section to the Xray core editor with an “always on” view for required services.
  * Added interactive toggles for optional gRPC API services plus warnings/errors for unrecognized services.
  * Extended UI translations (EN/FA/RU/ZH) and added localized API copy for the core editor and configuration modal.
  * Improved header text with localized fallback title/description support.
* **Tests**
  * Added a Node test suite covering API service selection, validation, and config updates.
* **Chores**
  * Updated Node.js engine requirement and adjusted ignore rules for local build artifacts.


ImMohammad20000:
@coderabbitai review

Multi-Engineer-dev:
**1. Unpinned Node for the test script — applied (441c1796), via an `engines`
pin rather than switching to `bun test`.** Pinned `node >= 22.18.0`, which is
the actual floor: the script runs `node --test` over `.ts` files, relying on
native type stripping (default since 22.18/23.6) and test-runner globs. I
deliberately did NOT move to Bun's runner: the tests import `node:test` /
`node:assert/strict`, and Bun's node:test compatibility is a partial mapping
onto bun:test — keeping the Node runner and documenting the constraint is the
minimal deterministic fix.

**2/3. Missing locale keys — applied (81fadea9), and extended.** Added
`coreEditor.api.alwaysOn` / `.unknown` to all four locales (en included, for
key-parity). While fixing it I found the same gap in the legacy modal, which
this review didn't flag: `coreConfigModal.apiServices`, `.apiServiceAlwaysOn`,
and `.apiServicesUnknown` were also read with inline defaults and missing from
every locale — added those too. Two wording deviations from the suggested
strings: fa uses «همیشه فعال» (matching the hint's own فعال terminology, rather
than «روشن»), and ru uses «всегда включён» (agreeing with «сервис», rather than
the impersonal «включено»).

**4. Xray-only validation applied to WG configs — applied (1acea754).** The
render was already gated on `isXrayBackend`; the submit check ran before the
backend type was even read. The unknown-services check now sits behind
`backendType !== 'wg'` and the derivation memo returns empty lists for WG
(dep added).

**5. Non-object JSON toggle guard — applied (1acea754), with a corrected
rationale.** The review's stated failure ("passing null … throws when it
spreads the config") doesn't exist: object spread of `null` is legal JavaScript
(`{ ...null }` → `{}`), and I verified there is no throwing path anywhere in
`setOptionalService`. What the guard actually fixes is quieter and real:
toggling on valid-but-non-object JSON silently replaced the config (null /
primitives) or mangled it (an array becomes an index-keyed object). So the
guard is in as hardening against silent config corruption — not a crash fix,
since none existed.

**6. `setOptionalService` accepting required services — applied (dc40d39c).**
Lookup is now restricted to `OPTIONAL_API_SERVICES`, with enable- and
disable-no-op tests pinning the invariant (same-reference return). For the
record this was unreachable from the UI (both call sites only pass optional
entries) and harmless even if reached (the node's `sanitizeAPIServices` dedupes
required names) — it's invariant hygiene for the exported helper.

**Not addressed:** the docstring-coverage pre-merge warning (35.71% < 80%).
That's a repo-wide checker default, not an actionable finding on this diff —
the dashboard codebase doesn't follow a docstring convention for TS modules,
and padding these files with comment blocks to satisfy a threshold would cut
against the existing style.

ImMohammad20000:
@coderabbitai review

Multi-Engineer-dev:
Fixed in 7da71806 — with two notes.

**Scope:** this placeholder isn't from this branch. `git blame` traces
`zh.json`'s `"disabled": "??"` to e012c4f1 (“feat(api_keys): … (#549)”) on
`dev`; it entered this review's window via the incremental diff across the
rebase. Fixed here anyway since it's a one-line locale correction, but flagging
the provenance so it isn't attributed to this PR's changes.

**The finding was also incomplete:** the same `admins.disabled` key carried the
same placeholder in fa (`"???????"`) and ru (`"????????"`) — English was the
only locale with a real value. All three are now filled in, each matching the
locale's existing “disabled” terminology elsewhere in the file: zh `已禁用`,
fa `غیرفعال`, ru `Отключен`.

Validated: all four locale files parse and no `?`-placeholder values remain on
the key.

PR #672, opened 2026-07-03
feat(wireguard): support WireGuard as an Xray inbound

bitwiresys:
## What
Lets the panel treat a `wireguard` inbound **inside an Xray core** as a first-class inbound: its tag is exposed (so users can be assigned and subscriptions generated for it), and its per-user traffic / online state are recorded. Purely additive — the standalone WireGuard backend (`CoreType.wg`) is untouched.

## Why
Xray now ships a first-class WireGuard **inbound with a UserManager** (XTLS/Xray-core#6360, in v26.6.27). That makes it possible to serve WireGuard through the same Xray core as VLESS/VMess/Trojan/Shadowsocks — gaining Xray routing, sniffing, unified stats, and transport obfuscation (finalmask) for WG clients — as an alternative to the kernel WG backend. For the panel to use it, the Xray config reader must recognize a `wireguard` inbound, and usage recording must map WG stats back to panel users.

## How
- `app/core/xray.py` — add `_read_wireguard_inbound`, dispatched from `_read_inbound` when `protocol == "wireguard"`. Surfaces the inbound tag (appears in `/api/inbounds`) and captures public params (listen port, address, mtu, derived public key).
- `app/jobs/record_usages.py` — Xray's WG inbound reports per-user stats keyed by the peer's **hex public key** (`user>>>{hex_pubkey}>>>traffic`) rather than the numeric `{id}`. Add a cached `{hex_public_key -> user_id}` map and translate those stat names before recording, so WG traffic and the online indicator attribute to the right user. Non-WG stats unaffected (the map is only consulted for non-numeric stat names).

## Compatibility / notes
- Additive; no schema change, native WG backend untouched.
- Pairs with the node-side change that provisions peers on the Xray WG inbound: PasarGuard/node#64.

## Testing
- `ruff check` passes on the changed files; `py_compile` clean.
- WG-in-Xray verified end-to-end against a live node: user connects, `used_traffic` climbs and `online_at` updates each cycle; 15-minute soak, no disconnects, no client/server errors.

## Summary by CodeRabbit

* **New Features**
  * Added WireGuard support in Xray inbounds, including automatic handling of public keys, addresses, MTU (defaulting to 1420), and UDP networking.
  * WireGuard inbounds are now properly included and tagged in configuration editing.

* **Bug Fixes**
  * Improved WireGuard inbound preparation to default missing pre-shared keys to an empty value.
  * Enhanced usage statistics processing to better handle malformed responses.

* **UI/Validation**
  * Updated inbound form validation and rendering rules for WireGuard inbounds during edit/commit.

* **Sanitization**
  * Adjusted WireGuard inbound sanitization to remove transport/security fields when applicable.


ImMohammad20000:
is ther any changes require in https://github.com/PasarGuard/node_bridge_py ?

bitwiresys:
> is ther any changes require in https://github.com/PasarGuard/node_bridge_py ?

no change required

bitwiresys:
Synced with upstream main and dropped the WireGuard stats workaround, see [PasarGuard/node#64](https://github.com/PasarGuard/node/pull/64) for the root cause.

`record_usages.py` kept a hex-pubkey-to-user-id cache to translate stats keys for the WireGuard inbound, on the assumption xray reports WG peers by their public key. But the node already sends `str(user.id)` as Email for WireGuard users, same as vmess/vless/trojan/shadowsocks/hysteria. Once both the xray-core side and node side stop overriding Email with the pubkey, WireGuard stats arrive keyed by the plain id like every other protocol. The mapping had nothing to attach to, it just silently dropped WireGuard stats.

Also fixed a Python 2-style `except ValueError, TypeError:` in `_process_users_stats_response` that would have crashed the job at runtime, and removed the unused `base64` import.

Verified end-to-end in Docker: panel + node + xray v26.7.11, real WireGuard client, traffic through the tunnel, used_traffic updates correctly on the user, no pubkey mapping involved.

ImMohammad20000:
@coderabbitai review

bitwiresys:
Rebased onto the latest `dev` (v5.1.0) and added fixes for the WireGuard inbound in the core editor:

- Edit dialog rendered the transport/security form instead of the WireGuard fields, because the editor hydrates `transport`/`security` onto the inbound and the branch was picked by key presence. Secret key, MTU, address and peers were unreachable after saving an inbound.
- `commitEditInbound` matched the same transport/security branch first, so `wgSecretKey` was never validated on edit.
- Saving a WireGuard inbound persisted the hydrated `transport`/`security` keys into the config; they are now stripped in `sanitizeProfileInbounds`.
- Auto-generated tag used the transport value and produced `WIREGUARD TCP …`; WireGuard is UDP-only, and the security segment is skipped for it.

Local CI is green: `ruff check`, `tsc --noEmit` and the dashboard build. Backend pytest shows the same pre-existing failures as a clean `dev` checkout.

bitwiresys:
One more fix: creating a host for a WireGuard inbound returned 500.

`_prepare_subscription_inbound_data` passed `inbound_config.get("pre_shared_key", None)` into `SubscriptionInboundData.wireguard_pre_shared_key`, which is typed `str`. A WireGuard inbound has no top-level `pre_shared_key` (PSK is per peer), so the value was `None` and pydantic raised a ValidationError on every WireGuard host creation. Since subscriptions are built from hosts, no WireGuard host could exist and the subscription came back empty.

ImMohammad20000:
@coderabbitai review

bitwiresys:
Synced this branch with the current official `dev` branch. Retained WireGuard over Xray support and the PSK bridge. Added bulk-expiry initialization for users with `expire = null`, which covers WireGuard subscriptions without an expiry.

PR #695, opened 2026-07-13
feat(core): add REALITY target scanner

parsa222:
## Summary

Adds a REALITY target scanner: a "Scan target" button in the Xray inbound
editor (next to a REALITY `dest`/`target` field) that probes a candidate
handshake target and reports whether it is a good REALITY decoy, plus a
`POST /api/core/reality-scan` endpoint behind it.

For a given host it checks:
- TLS 1.3
- HTTP/2 (ALPN)
- Key-exchange group (X25519, and post-quantum X25519MLKEM768)
- HTTP/3 advertisement (Alt-Svc)
- Certificate validity, issuer/subject, expiry, and the non-wildcard SANs
  (surfaced as suggested valid SNIs)
- Handshake latency

The verdict (`feasible`) requires the hard REALITY requirements: TLS 1.3 +
HTTP/2 + a valid certificate, and it is not vetoed unless a non-X25519 group
was definitively observed. Post-quantum and HTTP/3 are reported as extras and
never fail the scan.

## How it works

Three isolated probes, so a partial failure degrades one field to "unknown"
instead of failing the whole scan:
1. Main TLS probe via Python `ssl` (TLS version, ALPN/H2, certificate parsed
   with `cryptography`).
2. A self-contained raw TLS 1.3 ClientHello/ServerHello parser to detect the
   negotiated key group. This is done by hand because the panel's OpenSSL 3.0
   predates X25519MLKEM768 (OpenSSL 3.5), so it cannot rely on the local stack.
3. Best-effort HTTP/3 detection from the server's `Alt-Svc` header.

## Security

The endpoint could otherwise be an SSRF primitive, so:
- Every target is resolved and pinned to a public IP. Private, loopback,
  link-local, reserved and cloud-metadata addresses are rejected.
- Connections are pinned to the resolved IP and never re-resolved, which closes
  the DNS-rebinding hole.
- DNS resolution is time-bounded and concurrent scans are capped by a semaphore.
- The endpoint is gated behind the `cores:read` permission.

## API

`POST /api/core/reality-scan` -> `RealityScanResult`
Request: `{ target, sni?, timeout? }` (timeout clamped to 1-20s, default 10).

## Tests

Backend unit tests in `tests/test_reality_scan_unit.py` cover target parsing,
the SSRF public-host guard, the ClientHello/ServerHello group probe,
certificate/SAN parsing, and the feasibility logic (no network required; one
opt-in network integration test behind `REALITY_SCAN_NETWORK_TEST=1`).

## Notes

- **i18n: no locale keys are included.** All new UI strings use
  `t(key, { defaultValue: '...' })`, so they render in English and fall back
  cleanly with translations missing. Adding `fa`/`zh`/`ru` entries under
  `coreEditor.realityScan.*` (and `coreConfigModal.scanRealityTarget`) is a
  good follow-up.
- The dashboard uses a small hand-written client (`service/reality-scan.ts`)
  instead of the generated one, mirroring the generated call shape. It can be
  folded into the generated API client once codegen is re-run.

## Summary by CodeRabbit

* **New Features**
  * Added REALITY target scanning via a new Core API endpoint (`/api/core/reality-scan`), returning structured TLS/HTTP/feasibility and certificate metadata.
  * Introduced a “Scan target” action in the Xray editor with multi-target tokenized scanning, bounded parallel execution, per-target status tracking, result summaries, and expandable details.
  * Supports optional SNI override, configurable timeout, and scan cancellation/abort.
* **Tests**
  * Added comprehensive unit tests for target parsing, DNS/public-IP resolution, TLS handshake/client-hello parsing, certificate extraction, feasibility logic, concurrency limits, and an optional live network integration test.


ImMohammad20000:
@coderabbitai review

ImMohammad20000:
@coderabbitai review

parsa222:
@coderabbitai full review

parsa222 (commented):
next commit

ImMohammad20000:
@coderabbitai review

Kiya6955:
the scanner should test the IP and SNI together, not just the IP on its own. so like, take 1.1.1.1:443 with SNI example.com, try the handshake with those two together and check if it comes back TLS 1.3 and if the cert actually matches that SNI. if both work then that pair is good to use. also worth knowing a different SNI on the same IP might not work the same way, so each pair kinda needs its own test

parsa222:
> the scanner should test the IP and SNI together, not just the IP on its own. so like, take 1.1.1.1:443 with SNI example.com, try the handshake with those two together and check if it comes back TLS 1.3 and if the cert actually matches that SNI. if both work then that pair is good to use. also worth knowing a different SNI on the same IP might not work the same way, so each pair kinda needs its own test



yeah it already does exactly this. the whole point of reality is to mimic the TLS, it's not even about the ip, and that's what the scan actually checks. it does the tls 1.3 handshake with the sni you give it and validates the cert against that sni, so the sni is what decides it, not the ip. (the SAN output is just a cherry on top)

look at the two tests, same endpoint (1.0.0.1): as a bare ip with no sni it returns "not suitable" (nothing for the cert to match), but as one.one.one.one with the sni it returns "suitable". 



resolved : 



and you can check it yourself, same ip different sni: (get an api key or use the owner user/pass)

    POST /api/core/reality-scan  {"target":"1.0.0.1:443","sni":"www.google.com","timeout":10}
which returns:
    cert_valid=false, reason: Hostname mismatch, certificate is not valid for 'www.google.com'

in the UI it's resolving the ip from the target and checks the cert against the sni (target), and the results are based on that. that's the whole point.


another example: 



and you might be wondering, why H3 and post quantum then? 

the h3 check is a nice bonus if you're also picking a host for hysteria masquerade (that fronts a real http/3 site), and the post quantum check matters because modern browser + cloudflare negotiate x25519mlkem768 now, so it flags whether a decoy matches a current browser fingerprint and whether you'd want to add pq or not.
(pq check might have been an overkill but why not....)
https://xtls.github.io/en/config/transports/reality.html#realityobject

PR #712, opened 2026-07-20
fix: update download_settings host changes without re-saving the main host

FunLay123:
## Summary

When a host is used as another host's xhttp download_settings, editing that host alone did not update the main host's cached data. Now, saving a host also refreshes every host that points to it as their download_settings.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests / CI

## Checklist

### DIFF
--- a/app/operation/host.py
+++ b/app/operation/host.py
@@ -95,7 +95,13 @@ class HostOperation(BaseOperation):
         host = BaseHost.model_validate(db_host)
         asyncio.create_task(notification.modify_host(host, admin.username))
 
-        await host_manager.add_host(db, db_host)
+        db_hosts = await get_hosts(db=db)
+        dependents = [
+            h
+            for h in db_hosts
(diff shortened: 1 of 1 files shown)


M03ED:
its better to have a query param for `get_hosts` and only get hosts that have download settings we want to change instead of checking all (at least for PostgreSQL this is possible)

FunLay123:
@M03ED Fair point, but I don't think it's worth the added complexity here. Hosts aren't a high-cardinality table like users — realistically dozens, maybe a couple hundred on the largest setups — and this query only runs when an admin explicitly saves a host, not on the subscription-serving hot path. A SQL-level filter would also need separate handling per DB dialect. Happy to remake if host counts become a real problem, but for now I'd rather keep the simple way.

PR #723, opened 2026-07-26
fix: redirect authenticated users away from /login

iAliF:
The login page unconditionally cleared the auth token and query cache on every mount, forcing a re-login even when a valid session already existed. It now checks the stored token with getCurrentAdmin() (the same check the dashboard route loader uses) and redirects to the dashboard if it's still valid, only clearing the session when it's missing or invalid. The check is skipped for the Telegram MiniApp flow, which manages its own auth token independently.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests / CI

## Checklist

- [ ] I tested the change locally or explained why it cannot be tested.
- [X] I did not include secrets, tokens, private keys, or unrelated changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login-session behavior when navigating to and from the login page.
  * Automatically redirects already-authenticated users to the dashboard.
  * Safely validates any stored token before proceeding, with reliable handling of aborted requests.
  * Cleans up invalid/expired sessions (canceling active requests and clearing cached data) only when access is rejected.
  * Kept the existing Telegram MiniApp authentication flow unchanged.

### DIFF
--- a/dashboard/src/pages/login.tsx
+++ b/dashboard/src/pages/login.tsx
@@ -8,9 +8,9 @@ import { Input } from '@/components/ui/input'
 import { LoaderButton } from '@/components/ui/loader-button'
 import { PasswordInput } from '@/components/ui/password-input'
 import { Tabs, TabsList, TabsTrigger } from '@/components/ui/tabs'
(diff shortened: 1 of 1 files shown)


iAliF:
@coderabbitai review

PR #772, opened 2026-08-14
perf(groups): scale membership summaries

### LINKED ISSUE #771: Scale group summaries for large user counts
## Problem

`GET /api/groups` currently eager-loads every `User` row attached to every returned group only to compute `total_users` with `len(group.users)`. On installations with many users this makes response time and memory use grow with the number and width of user records, even though the response only needs a count.

The association-table primary keys are ordered `(user_id, groups_id)` and `(inbound_id, group_id)`. Group-centric counts, joins, and bulk deletes therefore lack a deterministic reverse access path on databases that do not automatically index foreign-key columns.

## Evidence

Synthetic local SQLite benchmark with 500,000 user memberships across 100 groups:

| Path | Time | Traced Python allocations | Query plan |
| --- | ---: | ---: | --- |
| Current full user hydration | 1,014.19 ms | 74.43 MiB | Loads user rows |
| SQL count, no reverse index | 2,196.56 ms median | — | Full association scan |
| SQL count with covering reverse index | 18.016 ms median | 0.0098 MiB peak | Covering-index search |

This benchmark is directional rather than a production guarantee, but it reproduces both scaling problems: object hydration and reverse membership lookup.

## Proposed change

- Populate `Group.total_users` with a correlated SQL count expression in the group-list query.
- Keep the `users` relationship unloaded for group summaries.
- Add covering reverse indexes:
  - `users_groups_association (groups_id, user_id)`
  - `inbounds_groups_association (group_id, inbound_id)`
- Build/drop the PostgreSQL indexes with `CONCURRENTLY` outside the migration transaction so membership writes remain available during upgrades.
- Preserve existing create/update/single-group behavior by setting `total_users` whenever those paths intentionally load the relationship.

## Acceptance criteria

- `/api/groups` returns the same `total_users` values without loading `Group.users`.
- Group-centric membership queries use the new reverse indexes.
- Upgrade and downgrade work on SQLite, PostgreSQL, and MySQL/MariaDB.
- PostgreSQL production upgrades do not use a write-blocking normal index build.
- Regression tests cover the index definitions and unloaded relationship behavior.

dr-hoseyn:
## Summary

- calculate `Group.total_users` in SQL for the group-list query instead of hydrating every related `User`
- add covering reverse indexes for group-centric user/inbound membership lookups and bulk deletes
- build and drop the PostgreSQL indexes concurrently to keep membership writes available during upgrades
- preserve create/update/detail response behavior and add regression coverage

Closes #771

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Not applicable: API schema is unchanged.)

## Testing

- `uv run ruff check .` — passed
- `uv run ruff format --check app/db/models.py app/db/crud/group.py tests/test_group_membership_indexes.py tests/api/test_group.py` — passed
- `uv run pytest -q tests/test_group_membership_indexes.py tests/api/test_group.py::test_groups_get_counts_users_without_loading_user_rows` — 3 passed
- relevant group/bulk API suite — 18 passed
- `uv run alembic check` on SQLite — no new upgrade operations
- empty SQLite database upgrade to head, index introspection, and one-revision downgrade — passed
- PostgreSQL offline SQL compilation — emits `COMMIT`, `CREATE INDEX CONCURRENTLY`, then `BEGIN`; downgrade emits `DROP INDEX CONCURRENTLY`
- MySQL and SQLite migration SQL compilation — passed

Synthetic SQLite benchmark (500,000 memberships / 100 groups):

- current full user hydration: 1,014.19 ms and 74.43 MiB traced Python allocations
- SQL summaries without reverse index: 2,196.56 ms median (association scan)
- SQL summaries with covering reverse index: 18.016 ms median and 0.0098 MiB peak (covering-index search)

The benchmark is directional and is not presented as a production guarantee.

## Screenshots

Not applicable (backend/database-only change).

## Notes for reviewers

- The response schema and route behavior are unchanged.
- PostgreSQL concurrent index creation must run outside a transaction; the migration uses Alembic's `autocommit_block()` for this path.
- MySQL/MariaDB may create an implicit foreign-key index. The explicit composite definition gives the schema a deterministic covering index order for group-centric access; InnoDB can use it for foreign-key enforcement as well.
- SQLite uses normal index DDL, matching its typical single-node deployment model.

## Summary by CodeRabbit

* **Improvements**
  * Group listings now display accurate member totals without loading individual member records.
  * Group retrieval is more efficient, especially for groups with many members.
  * Reverse membership lookups are faster through improved database indexing.

* **Bug Fixes**
  * Corrected group member-count handling when group details are loaded incrementally.

* **Tests**
  * Added coverage to verify member totals, relationship loading behavior, and membership index configuration.


dr-hoseyn:
@coderabbitai review

PR #774, opened 2026-08-14
perf(users): avoid full-table sorts in list queries

### LINKED ISSUE #773: Avoid full-table sorts in paginated user lists
## Problem

The dashboard requests `/api/users` with the default `sort=-created_at`. The endpoint builds its total count from the already-ordered, wide ORM statement:

```sql
SELECT count(*)
FROM (
    SELECT <all user columns>
    FROM users
    ORDER BY users.created_at DESC
)
```

There is no global index beginning with `users.created_at`; the existing `(admin_id, created_at)` index only helps admin-scoped queries. On large installations, an unfiltered dashboard request can therefore sort/scan the full users table twice: once for the count and once for the paginated rows.

The page ordering also lacks a unique tie-breaker, so offset pagination is not deterministic when users share the same sort value.

## Evidence

SQLAlchemy compilation preserves the unnecessary `ORDER BY` for SQLite, PostgreSQL, and MySQL.

Synthetic in-memory SQLite benchmark with 500,000 users and a 50-row page:

| Operation | Current | Optimized | Current plan |
| --- | ---: | ---: | --- |
| Total count | 203.143 ms median | 0.224 ms median | full scan + temporary B-tree sort |
| Default page | 44.950 ms median | 0.052 ms median | full scan + temporary B-tree sort |

Adding the index alone still leaves the old count at 43.35 ms because it walks the ordered subquery. Both the count rewrite and the index are needed. This synthetic benchmark is directional, not a production latency guarantee.

## Proposed change

- Rewrite the total query as a narrow `count(users.id)` while preserving filters and joins and explicitly removing page ordering.
- Add `id` as a direction-matched tie-breaker for deterministic offset pagination.
- Add a `(created_at, id)` index for the dashboard's default order.
- Build/drop the PostgreSQL index concurrently so upgrades keep user writes available.
- Add regression tests for generated SQL, owner filters, stable ordering, and index metadata.

## Acceptance criteria

- The count query contains no `ORDER BY` and selects no wide user payload columns.
- Owner/admin filters still produce accurate totals.
- Default sorting is `created_at DESC, id DESC`.
- The schema exposes `(created_at, id)` in that order.
- Upgrade/downgrade works on SQLite, PostgreSQL, and MySQL/MariaDB.
- Existing user API tests remain green.

dr-hoseyn:
## Summary

- remove page ordering and wide user columns from the `/api/users` total-count query
- add a direction-matched `id` tie-breaker for deterministic offset pagination
- add `(created_at, id)` for the dashboard's default user-list order
- build/drop the PostgreSQL index concurrently to keep user writes available during upgrades
- add cross-dialect SQL and index regression tests

Closes #773

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Not applicable: API schema is unchanged.)

## Testing

- `uv run ruff check .` — passed
- targeted Ruff format check — passed
- `uv run pytest -q tests/test_user_list_query_scaling.py` — 4 passed
- targeted endpoint/count tests — 6 passed
- `uv run pytest -q tests/api/test_user.py` — 82 passed
- `uv run alembic check` on SQLite — no new upgrade operations
- empty SQLite migration chain to head — passed
- one-revision downgrade/upgrade and index introspection — passed
- PostgreSQL offline SQL — `CREATE INDEX CONCURRENTLY` / `DROP INDEX CONCURRENTLY`
- MySQL migration compilation — passed

Synthetic SQLite benchmark with 500,000 users and a 50-row page:

| Operation | Before | After |
| --- | ---: | ---: |
| Total count | 203.143 ms median | 0.224 ms median |
| Default page | 44.950 ms median | 0.052 ms median |

Before the change both plans used a full scan and temporary B-tree sort. The optimized page uses `idx_users_created_at_id`; the count no longer sorts. The benchmark is directional, not a production guarantee.

## Screenshots

Not applicable (backend/database-only change).

## Notes for reviewers

- Response fields, filters, and pagination parameters are unchanged.
- Owner/admin joins remain in the count statement; regression coverage verifies this.
- PostgreSQL concurrent DDL is isolated with Alembic's `autocommit_block()`.
- This PR and #772 both independently target the current `dev` migration head. If one lands first, the other migration should be rebased onto the new head before merge to avoid parallel Alembic heads.

## Summary by CodeRabbit

* **Performance**
  * Improved user list loading and counting, especially for filtered results.
  * Added a database index to support faster, more consistent user sorting.

* **Bug Fixes**
  * User lists now maintain stable ordering when multiple users share the same creation time.
  * User counts preserve existing filters and owner-related conditions accurately.


dr-hoseyn:
@coderabbitai review

PR #776, opened 2026-08-14
perf(users): eliminate refresh storms in status jobs

### LINKED ISSUE #775: Avoid per-user refresh storms in scheduled status transitions
## Problem

The 30-second expiry and data-limit jobs bulk-update matching users, then call `refresh_and_load_user()` once per user. Each refresh reloads the user plus eager relationships, turning one status transition into roughly three extra SELECTs per user.

A burst of expiring or limited accounts therefore creates a linear query storm while the scheduler uses `max_instances=1`, delaying later job runs on large installations.

## Reproduction / measurement

On SQLite with 200 eager-loaded users:

- current path: 601 statements (1 UPDATE + 600 SELECTs), about 646 ms
- bulk update without per-user refresh: 1 statement, 0 SELECTs, about 22 ms

These are synthetic local measurements, but the query-count growth is deterministic (`3N + 1` in this setup).

## Proposed fix

- Keep the existing single bulk UPDATE.
- Update `status` and `last_status_change` on the already eager-loaded ORM objects in memory.
- Do not refresh each object after the commit.
- Return early for an empty user list.
- Add a regression test that counts SQL statements and verifies preloaded relationships remain available.

No schema migration or API change is required.

dr-hoseyn:
## Summary

The scheduled expiry and data-limit jobs already load each user's required relationships before performing a bulk status update. This change removes the subsequent per-user refreshes, which generated about three SELECTs per user during large transition bursts.

The returned ORM objects are updated in memory after the successful commit, preserving the existing downstream behavior without reloading unchanged relationships. Empty batches now return without issuing SQL.

Closes #775

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Not applicable)
- [x] I checked database migrations when models or schema changed. (No schema change)

## Testing

- `uv run ruff check app/db/crud/user.py tests/api/test_user.py`
- `uv run pytest tests/api/test_user.py -q` — 83 passed
- Regression test observes exactly one UPDATE and zero SELECTs while transitioning three users, and verifies eager-loaded groups and usage logs remain present.

Synthetic SQLite measurement for 200 eager-loaded users:

- Before: 601 statements (1 UPDATE + 600 SELECTs), ~646 ms
- After: 1 statement (1 UPDATE + 0 SELECTs), ~22 ms

## Screenshots

Not applicable.

## Notes for reviewers

No migration or API change. The optimization is limited to `update_users_status()`, currently used by the scheduled expiry and data-limit transitions. Sessions are configured with `expire_on_commit=False`, and the selected users have their downstream relationships eager-loaded before this function is called.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bulk user status updates for faster processing.
  * Preserved related user information during status changes.
  * Handled empty user lists without unnecessary database work.
  * Ensured users receive a consistent status-change timestamp during bulk updates.

* **Tests**
  * Added coverage for efficient bulk updates, empty-list handling, timestamp consistency, and preservation of loaded user details.


dr-hoseyn:
@coderabbitai review

dr-hoseyn:
@coderabbitai review

dr-hoseyn:
@coderabbitai full review

dr-hoseyn:
Addressed the full-review test hardening in dabd3e1a: persisted status and the shared last_status_change are now verified through an independent TestSession after SQL capture is removed, and listener/user cleanup is failure-safe with rollback plus deletion in finally. Ruff passes, the focused regression passes, and the full tests/api/test_user.py suite passes (83 tests).

@coderabbitai review

dr-hoseyn:
@coderabbitai full review

PR #780, opened 2026-08-14
ops(docker): rotate panel container logs

### LINKED ISSUE #779: Bound Docker JSON logs to prevent disk exhaustion
## Problem

The deployment Compose file uses Docker's default `json-file` logging without size or file limits. On a busy installation this can consume disk continuously and eventually threaten PostgreSQL and the whole host.

In an anonymized production observation, the panel container log reached approximately 2.09 GB and 11.7 million lines in about two days.

## Proposed change

Set bounded JSON logging for the panel service:

- `max-size: 25m`
- `max-file: 4`

This caps retained panel container logs at roughly 100 MB per container while preserving recent diagnostics.

## Live validation

The same limits were applied to a production installation and only the panel container was recreated. It returned healthy, approximately 2 GB of disk was recovered, and the database/connection-pool containers were not restarted.

## Operational note

The setting takes effect when the container is recreated. Recreating a container removes its previous JSON log, so operators should archive it first if retention is required.

dr-hoseyn:
## Summary

- bound the panel container's Docker JSON logs to four 25 MB files
- prevent unbounded log growth from exhausting disk on busy installations

Closes #779

## Validation

- parsed `docker-compose.yml` with PyYAML and asserted the resulting logging configuration
- `git diff --check`
- production canary with the same settings: panel returned healthy after a controlled single-service recreation and approximately 2 GB of disk was recovered

## Operational note

The limits take effect when the panel container is recreated. Operators that need the existing Docker JSON log should archive it before recreation.

## Summary by CodeRabbit

* **Chores**
  * Improved application log management by limiting log files to 25 MB and retaining up to four files.
  * Helps prevent excessive disk usage from accumulated logs.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,11 @@ services:
   pasarguard:
     image: pasarguard/panel:latest
     restart: always
+    logging:
+      driver: json-file
+      options:
+        max-size: "25m"
+        max-file: "4"
     env_file: .env
     network_mode: host
     volumes:


dr-hoseyn:
@coderabbitai review

PR #791, opened 2026-08-14
perf(system): collapse dashboard user metrics into one query

### LINKED ISSUE #790: perf(system): collapse dashboard user metrics into one query
## Problem

`GET /api/system/users` builds dashboard user metrics with two sequential database statements: one grouped status count and one recent-online count. The dashboard polls this endpoint every five seconds and the Statistics page every two seconds, so every active dashboard continuously multiplies database round trips.

The current queries return correct data and use admin-scoped indexes where available, but the two scans/round trips are avoidable because all metrics come from the same filtered `users` relation.

## Proposed solution

- Compute total, per-status, and recent-online counts with one conditional-aggregation statement.
- Preserve global and admin-scoped behavior, the two-minute online window, and the existing API response.
- Keep the result fully fresh; do not introduce a cache or stale response window.
- Add regression coverage for metric values, admin isolation, empty datasets, and a one-SELECT query budget.
- Validate generated SQL on supported dialects and benchmark the consolidated query against the current two-query path.

## Alternatives considered

- Frontend polling changes: already covered by #604 and do not remove backend duplication for visible dashboards.
- Short-lived caching: reduces load but introduces multi-worker invalidation/freshness concerns and is not required to remove the duplicate round trip.
- New indexes: should only be added if production-representative `EXPLAIN` output shows that the consolidated query needs them.

## Area

API / Database

## Impact

Important. This is a continuously polled dashboard endpoint, so small per-request savings compound across active administrators and workers.

## Acceptance criteria

- One SQL SELECT returns all user count metrics for the endpoint.
- Global and admin-scoped metrics match current behavior.
- Empty status groups return zero.
- Online users use the existing two-minute threshold.
- No API schema, cache, migration, or frontend change.
- Focused tests and a reproducible benchmark are included in the PR evidence.

## Version

Current `dev` at `3331421e`.

dr-hoseyn:
## Summary

- Collapse dashboard status counts and recent-online count into one database statement.
- Use conditional aggregation for global metrics and index-friendly scalar aggregates for admin-scoped metrics.
- Preserve the existing API response, two-minute online window, and admin visibility rules.
- Add correctness, query-budget, and SQLite/PostgreSQL/MySQL compilation coverage.

Closes #790

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

No documentation or migration is needed because the endpoint contract and schema are unchanged.

## Testing

- `.venv\Scripts\python.exe -m ruff check .` — all checks passed
- `.venv\Scripts\python.exe -m pytest -q tests/api/test_system.py tests/test_system_user_metrics.py` — 10 passed
- `.venv\Scripts\python.exe -m pytest -q` — 532 passed, 2 skipped

Directional SQLite benchmark with 500,000 synthetic users and indexes matching the model:

| Scope | Before (2 statements) | After (1 statement) | Change |
| --- | ---: | ---: | ---: |
| Global | 189.471 ms | 100.954 ms | -46.7% |
| Admin-scoped | 0.435 ms | 0.180 ms | -58.7% |

These local synthetic numbers are not a production latency guarantee; they validate the query shape and guard against the scoped conditional-aggregation regression discovered during implementation.

## Screenshots

Not applicable (backend-only change).

## Notes for reviewers

The global and scoped SQL shapes are intentionally different. A single conditional scan was faster globally, but it regressed the scoped path by bypassing the existing composite indexes. Scalar aggregate subqueries keep the scoped request to one round trip while allowing each count to use `(admin_id, status)` or `(admin_id, online_at)`.

Frontend polling changes are intentionally out of scope.

## Summary by CodeRabbit

- **New Features**
  - User metrics now include per-status counts, total users, and users active within the last two minutes.
  - Metrics can be filtered to a specific administrative scope.
  - Empty scopes and all supported user statuses are handled consistently.

- **Performance**
  - User metrics are collected through a consolidated query for more efficient system reporting.

- **Tests**
  - Added coverage for global and scoped metrics across supported database systems.

### DIFF
--- a/app/db/crud/user.py
+++ b/app/db/crud/user.py
@@ -864,7 +864,11 @@ async def get_users_count_metrics(
     online_window: timedelta,
     admin_id: int | None = None,
 ) -> tuple[dict[str, int], int]:
-    """Return per-status, total, and recent-online user counts in one SELECT."""
+    """Return per-status, total, and recent-online user counts in one SELECT.
+
+    The ``total`` value includes only the requested statuses. Pass every
+    ``UserStatus`` value when a complete user count is required.
+    """
     stmt = _build_user_count_metrics_query(statuses, datetime.now(UTC) - online_window, admin_id)
     row = (await db.execute(stmt)).one()
 
--- a/app/db/crud/user.py
+++ b/app/db/crud/user.py
@@ -827,37 +827,50 @@ async def get_users_count(db: AsyncSession, status: UserStatus = None, admin_id:
     return result.scalar()
 
 
-async def get_users_count_by_status(
-    db: AsyncSession, statuses: list[UserStatus], admin_id: int | None = None
-) -> dict[str, int]:
-    """
-    Gets count of users grouped by status in a single query.
+def _build_user_count_metrics_query(
+    statuses: list[UserStatus], online_since: datetime, admin_id: int | None = None
+) -> Select:
+    """Build one index-aware query for dashboard user counts."""
+    if admin_id is not None:
+        status_columns = [
+            select(func.count(User.id))
+            .where(User.admin_id == admin_id, User.status == status)
+            .scalar_subquery()
+            .label(status.value)
+            for status in statuses
+        ]
+        online_column = (
+            select(func.count(User.id))
+            .where(
+                User.admin_id == admin_id,
+                User.online_at.isnot(None),
+                User.online_at >= online_since,
+            )
+            .scalar_subquery()
+            .label("online")
+        )
+        return select(*status_columns, online_column)
 
-    Args:
-        db (AsyncSession): Database session.
-        statuses (list[UserStatus]): List of statuses to count.
-        admin_id (int, optional): Filter by admin.
-    Returns:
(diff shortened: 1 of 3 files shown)


dr-hoseyn:
@coderabbitai review

dr-hoseyn:
Addressed the actionable review feedback in `0249bf59`:

- documented that `total` is the sum of the requested statuses;
- extracted the seeded async fixture and split correctness/scoping cases into focused tests;
- expanded the one-statement regression test to cover both global and admin-scoped paths;
- re-ran project-wide Ruff and the focused suite: 11 passed.

I kept the admin-scoped scalar aggregate shape intentionally. The suggested filtered conditional aggregation was benchmarked during implementation and regressed the scoped SQLite case from 0.403 ms to 2.596 ms (~6.4x), while the scalar-subquery statement measured 0.180 ms and preserves use of the existing `(admin_id, status)` and `(admin_id, online_at)` indexes.

dr-hoseyn:
@coderabbitai full review

PR #793, opened 2026-08-15
fix(nats): prevent KV scan consumer leaks

dr-hoseyn:
## Summary

- replace full-bucket `kv.keys()` scans with prefix-scoped metadata watchers
- always stop and explicitly delete the temporary JetStream consumer on both success and failure
- preserve the in-memory KV behavior used by tests and single-process fallbacks

This prevents large shared KV buckets from turning each per-node queue lookup into a full scan. It also avoids abandoned consumers accumulating after request timeouts, which otherwise creates a feedback loop of higher NATS load, sync delays, and subscription timeouts.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Not applicable)
- [x] I checked database migrations when models or schema changed. (No schema changes)

## Testing

- `python -m uv run --frozen pytest tests/test_nats_node_memory.py tests/test_node_manager_sync.py tests/test_connect_concurrency.py tests/test_record_usages.py -q` — 24 passed
- `python -m ruff check app/nats/kv_cas.py tests/test_nats_node_memory.py` — passed
- validated against a live multi-worker deployment: queues drained, transient consumer count returned to zero, and subscription requests completed successfully

## Screenshots

Not applicable.

## Notes for reviewers

`KeyValue.Watcher.stop()` only unsubscribes the client; nats-py leaves the ephemeral consumer until its inactivity threshold. The explicit consumer deletion in `finally` makes cleanup immediate and is covered for both normal completion and exceptions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved key listing for NATS key-value stores by applying prefix filtering server-side.
  * Excluded deleted entries from key listings.
  * Ensured watcher resources are cleaned up reliably, including when errors occur.
* **Tests**
  * Added coverage for filtering, cleanup, and error handling during key listing.


dr-hoseyn:
@coderabbitai review

PR #794, opened 2026-08-16
fix(nodes): prevent cross-worker lifecycle races

dr-hoseyn:
## Summary
- let sibling workers replace their local node controller without stopping the remote node
- treat lifecycle lease contention as a no-op after attach fails, so a transient 409 cannot overwrite the lease owner's result
- add regression coverage for sibling upsert/connect handling, local-only controller replacement, and lifecycle contenders

## Root cause
In multi-worker deployments, a worker receiving an `upsert` or `connect` synchronization event replaced its local controller and remotely stopped the node. That raced the origin worker's successful start. Concurrent contenders then persisted `Node lifecycle operation already in progress` over the owner's eventual result.

## Verification
- `ruff check .`
- `pytest tests/test_node_manager_sync.py tests/test_node_manager.py tests/test_connect_concurrency.py -q` (13 passed)
- `pytest --ignore=tests/api --ignore=tests/test_review_admins_unit.py -q` (156 passed, 2 skipped)
- two-worker production canary: the node reached `connected`, lifecycle settled at `healthy` with no active lease, and no node-409 entries recurred after deployment

## Production observation
About 23 minutes after the successful canary, the remote node VPS itself became unreachable on SSH and both node ports from the panel and an independent network path. The panel and the other nodes remained healthy, and the fixed lifecycle 409 did not recur. The later `Request timed out` state is therefore tracked as a separate infrastructure outage rather than evidence of the cross-worker race returning.

## Test-suite note
The repository contribution guide names `next`, but that branch is not present on the upstream remote, so this PR targets `dev`. The full API suite is currently blocked by its local test database/auth seeding setup; the affected node lifecycle unit and integration-adjacent tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node synchronization across workers without triggering unintended remote shutdowns.
  * Prevented concurrent connection attempts from overwriting the result produced by the worker that owns the operation.
  * Improved replacement of locally managed node controllers while preserving their connection state.

* **Tests**
  * Added regression coverage for concurrent connections, controller replacement, and synchronized node updates.


dr-hoseyn:
@coderabbitai review

ImMohammad20000:
i fixed thus myself

PR #806, opened 2026-08-21
fix: ECH for mihomo and sing-box

mdjvd:
## Summary

FYI: This is AI-written code
We've had ECH for Xray, but not for mihomo, and the implementation for Sing-Box was imperfect and unusable.
This PR fixes the ECH issue for both subscription formats.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Tests / CI

## Checklist

## Testing

I have tested it on both formats, and they work with no issues as far as I can see.

### DIFF
--- a/app/subscription/clash.py
+++ b/app/subscription/clash.py
@@ -251,6 +251,12 @@ class ClashConfiguration(BaseSubscription):
             }
             if security == "reality" and tls_settings.get("publicKey")
             else None,
+            "ech-opts": {
+                "enable": True,
+                "config": tls_settings.get("echConfigList"),
+            }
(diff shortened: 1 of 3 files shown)


mdjvd:
If the ech is a base-64 encoded config, there is no problem. But if we choose to use the other supported format by Xray (eg, "example.com+https://1.1.1.1/dns-query"), then Sing-box and Mihomo will not be able to start the profile and proxies at all.
So I think we have to have different fields for each format in hosts.

PR #821, opened 2026-08-28
fix(node): serialize full user sync against reconnect/update_node

ImMohammad20000:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when synchronizing users to connected nodes.
  * Prevented node replacements or reconnect checks from interrupting an in-progress full user synchronization.
  * Added clearer handling when a node is no longer connected during synchronization.


ImMohammad20000:
@coderabbitai review

PR #822, opened 2026-08-28
feat(hosts): add Xray sockopt domain strategy

TaraRostami:
## Summary

Adds a per-host `xray_sockopt_domain_strategy` option to control `streamSettings.sockopt.domainStrategy` on generated Xray JSON proxy outbounds.

The default is `AsIs` and remains implicit for backward compatibility, so existing hosts keep the same generated Xray configuration unless a different strategy is selected.

This is useful when a generated Xray client configuration needs Xray's built-in DNS resolver to resolve the outbound server hostname, while keeping the hostname itself in the Host configuration and other subscription formats.

The setting merges with existing Fragment/Noise `dialerProxy` configuration and is applied to generated VMess/VLESS/Trojan/Hysteria outbounds, Shadowsocks, and XHTTP download-side stream settings.

WireGuard keeps its existing protocol-specific `settings.domainStrategy`, and non-Xray subscription formats such as share links remain unchanged.

> [!IMPORTANT]
> The migration currently uses `7c4bd5128e62` as its `down_revision`, which was the
> Alembic head of `dev` when this PR was prepared. If new migrations are merged into
> `dev` before this PR, this migration should be rebased onto the new head to avoid
> creating multiple Alembic heads.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

## Testing

Tested locally on Windows with a development SQLite database.

Backend / migration checks:

```bash
uv run alembic upgrade head
uv run alembic current
make check-migrations
make check
```

## Feature-specific tests
```bash
uv run pytest tests/test_xray_sockopt_domain_strategy.py -v
uv run pytest tests/api/test_user.py -k xray_subscription_applies_host_sockopt_domain_strategy -v
```

The dedicated regression tests cover:
- Hysteria `streamSettings.sockopt.domainStrategy`
- Shadowsocks' independent outbound builder
- WireGuard retaining its protocol-specific `settings.domainStrategy`
- API → database → subscription propagation
- `AsIs` remaining implicit in generated Xray JSON
- merging `domainStrategy` with Fragment/Noise `dialerProxy`

The panel was also run locally and the feature was manually tested through the Host UI and generated Xray subscription output.

## Screenshots
Screenshot 2026-08-28 135153

## Notes for reviewers
- `AsIs` is the default and is intentionally omitted from generated Xray JSON to preserve existing behavior.
- Non-AsIs values are emitted as `streamSettings.sockopt.domainStrategy`.
- Existing Fragment/Noise dialerProxy settings are merged rather than overwritten.
- XHTTP download-side stream settings use the selected strategy in Xray JSON output.
- WireGuard is intentionally excluded because it already uses the protocol-specific `settings.domainStrategy`.
- Share-link and other non-Xray subscription formats are unchanged.
- The database migration adds the new host column with `AsIs` as a non-null server default.


ImMohammad20000:
why sockopt domain strategy should be in hosts?

TaraRostami:
> why sockopt domain strategy should be in hosts?

Hello

Because `sockopt.domainStrategy` controls how Xray resolves the destination address of the generated proxy outbound, and that destination comes directly from the Host (`address`, `port`, transport, etc.).

The option itself is necessary to support configurations where the outbound server address must be resolved by Xray rather than left as-is. Without it, a Host whose address is a domain has no way to request Xray-side resolution in the generated Xray subscription. This is especially important for setups that rely on Xray's built-in DNS, including `dns.hosts`, or that need an explicit IPv4/IPv6 resolution strategy.

Making it per-host allows different hosts to use different resolution behavior. For example, one host may need `UseIP` so its domain is resolved through Xray's built-in DNS (including `dns.hosts`), while another host should keep the default `AsIs`.

A global setting would force the same behavior on every host, and putting it in the client template would not be a good fit because the proxy outbound is generated dynamically from Host data rather than owned by the template.

That's why I treated it as an endpoint/Host property, with `AsIs` as the default so existing hosts keep their current behavior.

I'd be interested to hear your thoughts on whether you see a better place for this setting in PasarGuard's data model.

PR #842, opened 2026-09-01
fix(runtime): scope trusted TLS to outbound clients

### LINKED ISSUE #841: [Bug]: pip_system_certs injects truststore globally, blocking uvicorn's event loop on TLS handshakes
### Before you submit

### PasarGuard version

v5.3.0

### Installation type

Docker Compose

### What happened?

On a production instance (~500 active users, 5 nodes, 4 uvicorn workers), the
panel progressively degrades after roughly 24–31 hours of uptime until it
becomes unusable and has to be restarted.

**Symptoms**

- After ~24–31 hours of uptime, all 4 workers climb to 50–70% CPU each and
  stay there, saturating all 4 cores.
- The panel UI becomes very slow, then unresponsive.
- Request latency degrades from ~15ms to 7,000–31,000ms for trivial endpoints
  such as `GET /api/user/<username>`.
- Eventually requests start failing with 503 and a misleading database error.
- Only a container restart recovers it; the cycle then repeats.

**Findings**

Profiling points to `pip_system_certs`, which the runtime image ships. It is
loaded via `pip_system_certs.pth` at interpreter startup and calls
`truststore.inject_into_ssl()` unconditionally (`bootstrap.py:22`,
`wrapt_requests.py:29`). There is no opt-out environment variable.

This replaces the process-wide default `ssl.SSLContext`. `truststore`'s
typical use case is client-side verification against the OS trust store, but
because the injection is global it also affects uvicorn's **server** socket.
Incoming TLS handshakes then go through truststore's memory-BIO path instead
of OpenSSL's native path, and block the asyncio event loop.

`pip_system_certs` exists to let pip trust an internal CA behind a corporate
MITM proxy. It appears to serve no purpose in the panel runtime.

**What was ruled out first**

Before profiling, the following were checked and found healthy:

- **Thread leak** — each worker held exactly 8 threads, stable.
- **NATS connection leak** — ~42 connections, stable across 15 minutes.
- **Database** — 44 of 151 connections in use, `SHOW FULL PROCESSLIST` showed
  only idle sessions, no slow queries, no lock waits. Disk writes were
  effectively zero during a 5-minute sample.
- **Node health** — no `BROKEN`/`recovered` flapping in the preceding 2 hours;
  node container logs showed only successful `GetStats`/`GetBackendStats`.
- **Outbound gRPC to nodes** — exactly 5 stable connections, so node
  connection pooling works correctly.

The MySQL errors appear to be a *symptom*, not the cause: with the event loop
blocked, requests could not acquire a DB connection in time.

**Note on affected deployments**

This instance has uvicorn terminating TLS directly (`UVICORN_SSL_CERTFILE` /
`UVICORN_SSL_KEYFILE`), with HAProxy in front in `mode tcp` SNI passthrough.
Deployments that terminate TLS at a reverse proxy would not hit this code
path, which may explain why it is not widely reported.

### Steps to reproduce

1. Run the panel with `UVICORN_SSL_CERTFILE` and `UVICORN_SSL_KEYFILE` set,
   so uvicorn terminates TLS itself (reverse proxy in TCP passthrough mode).
2. Put it under real load — roughly 500 active users plus a Telegram bot
   polling the API at ~40 req/min.
3. Let it run for 24+ hours without restarting.
4. Observe all uvicorn workers sitting at 50–70% CPU and response times in
   the multi-second range.
5. Profile a worker: `py-spy record --pid <worker_pid> --duration 45`
6. See `do_handshake (truststore/_api.py:97)` dominating the event loop thread.

### Expected behavior

`pip_system_certs` should not be present in the runtime image. If it is needed
at build time, it should not persist into the final layer.

Alternatively, truststore injection should be skipped when uvicorn terminates
TLS, so that incoming handshakes use OpenSSL's native path and never block the
event loop.

The panel should stay responsive under sustained load without requiring a
periodic container restart.

### Logs and error messages

```shell
No tracebacks — the failure is a performance collapse, not a crash.

Once the event loop is saturated, requests time out and surface as a
misleading database error:

WARNING: App-factory - Database unavailable while handling GET /api/user/xxx:
OperationalError: (2003, "Can't connect to MySQL server on '127.0.0.1' ()")
INFO: "GET /api/user/xxx HTTP/1.1" 503 Service Unavailable - 23961.26ms


Typical latencies during the degraded state:

"GET /api/user/xxx/sub_update?offset=0&limit=1 HTTP/1.1" 200 OK - 31342.28ms
"GET /api/system/users HTTP/1.1" 200 OK - 17736.21ms
"GET /api/users/usage?period=day HTTP/1.1" 200 OK - 11012.74ms
"GET /api/user/xxx HTTP/1.1" 200 OK - 13296.50ms


The same endpoints return in 10–30ms when the panel is healthy.

py-spy profile taken during a spike (45s, 35,991 samples):

event loop thread (asyncio/runners.py): 12.48% of all samples
└─ do_handshake (truststore/_api.py:97): 9.97% of all samples


The remaining frames (`_worker` 75%, `_recv` 12.5%) are idle threads waiting
on queues. Relative to the event loop thread itself, `do_handshake` accounts
for ~80% of the loop's time — which saturates the worker, since in an async
architecture all request handling flows through that single thread.
```

### Environment

OS: Ubuntu 24.04 (x86_64, 4 cores, 8 GB RAM)
Database: MySQL 9.7 (containerized, same host)
Python version: 3.14.7
Reverse proxy / CDN: HAProxy in `mode tcp` (SNI passthrough, no TLS termination)

Additional:
- Installation: Docker Compose
- 4 uvicorn workers
- Separate backend / scheduler / node-worker containers with NATS
- ~500 active users, 5 nodes

### Relevant configuration

```yaml
UVICORN_HOST = "127.0.0.1"
UVICORN_PORT = 10000
UVICORN_SSL_CERTFILE = "/var/lib/pasarguard/certs/<domain>/fullchain.pem"
UVICORN_SSL_KEYFILE = "/var/lib/pasarguard/certs/<domain>/privkey.pem"
UVICORN_SSL_CA_TYPE = "public"
UVICORN_WORKERS = 4
UVICORN_PROXY_HEADERS = True
UVICORN_FORWARDED_ALLOW_IPS = "127.0.0.1"

HAProxy frontend is `mode tcp` with SNI routing to 127.0.0.1:10000 —
no TLS termination at the proxy.
```

### Screenshots or videos

Flame graphs from py-spy, in order: before removing `pip_system_certs.pth`,
10 minutes after removal, and 3 hours after removal (all under identical
production load).

Image

Image

Image

### Additional context

**Verification**

Removing the `.pth` file and restarting the container:

    docker exec <container> rm \
      /code/.venv/lib/python3.14/site-packages/pip_system_certs.pth
    docker restart <container>

Re-profiled twice afterwards under identical production load:

| | before | +10 min | +3 hours |
|---|---|---|---|
| `do_handshake` | 9.97% | 0 | 0 |
| event loop thread | 12.48% | 16.55% | 16.56% |
| actual request handling | not visible | 0.27% | 0.18% |

`do_handshake` disappears entirely in both post-change profiles, and the
numbers are stable three hours apart — no sign of gradual re-accumulation.
Real request work (`run_asgi` → proxy_headers → FastAPI routing) and
SQLAlchemy query frames (0.11%) become visible for the first time, both at
negligible cost.

Latency returned to normal immediately and no certificate errors appeared in
the logs after removal.

The three attached flame graphs correspond to the columns above, in order.

The evidence strongly indicates this is the cause, though the degradation
cycle previously took 24–31 hours to manifest, so long-term verification is
still pending. I'll follow up once the instance has run a full cycle without
the spike returning.

Note that this workaround is not persistent  it is lost on any
`docker compose up -d` or image update, which is why a fix in the image itself
is needed.

dr-hoseyn:
## Summary
- remove `pip-system-certs` and its startup `.pth` hook so it can no longer replace `ssl.SSLContext` process-wide
- explicitly install Debian `ca-certificates` in the runtime image
- build outbound-only native TLS contexts that combine the OS trust store with `certifi`
- use those contexts for Telegram/Discord notifications, generic webhooks, and aiogram (including proxied sessions)
- add regression coverage for native server TLS, trusted client roots, aiohttp scoping, and direct/proxied Telegram sessions

## Why this preserves webhook TLS

`pip-system-certs` was originally added for Telegram, Discord, and webhook certificate verification. Removing it without a replacement could regress those integrations on some deployments.

This revision preserves trusted outbound HTTPS without calling `truststore.inject_into_ssl()`: client sessions receive their own native OpenSSL context, while uvicorn and every other server-side TLS consumer keep CPython's original `ssl.SSLContext`. The runtime image also installs its CA store explicitly, and `certifi` is loaded in addition to OS roots.

aiogram does not expose a public SSL-context constructor argument, so its connector configuration is set in a small subclass. Regression tests cover both its direct and proxy connector paths.

## Verification
- `pytest -q tests/test_ssl_context.py`: **5 passed**
- `pytest -q --ignore=tests/api` with a fresh in-memory test URL: **172 passed, 2 skipped**
- targeted Ruff check and format check: passed
- `compileall` for `app` and the SSL tests: passed
- outbound HTTPS using the new client context: PyPI returned HTTP 200
- after the request, `ssl.SSLContext.__module__` remains `ssl`

The local API suite requires the project's persistent migrated test database; the previous local database references a migration that is no longer present, while an in-memory URL does not share the schema across its test connections. The repository's database-migration CI remains the authoritative check for that suite.

Fixes #841

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,6 @@ dependencies = [
     "aiorwlock>=1.5.0",
     "typer>=0.27.2",
     "pasarguard-node-bridge>=0.9.1",
-    "pip-system-certs>=5.3",
     "nats-py>=2.15.0",
 ]
 
--- a/tests/test_ssl_context.py
+++ b/tests/test_ssl_context.py
@@ -1,0 +1,6 @@
+import ssl
+
+
+def test_ssl_context_is_not_globally_replaced():
+    """Keep server-side TLS on CPython's native OpenSSL implementation."""
+    assert ssl.SSLContext.__module__ == "ssl"
--- a/uv.lock
+++ b/uv.lock
@@ -827,7 +827,6 @@ dependencies = [
     { name = "nats-py" },
     { name = "packaging" },
     { name = "pasarguard-node-bridge" },
-    { name = "pip-system-certs" },
     { name = "psutil" },
     { name = "pydantic" },
     { name = "pydantic-settings" },
@@ -880,7 +879,6 @@ requires-dist = [
     { name = "nats-py", specifier = ">=2.15.0" },
     { name = "packaging", specifier = ">=26.3" },
     { name = "pasarguard-node-bridge", specifier = ">=0.9.1" },
-    { name = "pip-system-certs", specifier = ">=5.3" },
     { name = "psutil", specifier = ">=7.2.2" },
     { name = "pydantic", specifier = ">=2.13.5" },
     { name = "pydantic-settings", specifier = ">=2.15.0" },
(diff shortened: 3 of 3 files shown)


dr-hoseyn:
@coderabbitai review

ImMohammad20000:
This is not added for uvicorn ssl its added for webhooks ssl

dr-hoseyn:
Superseded by #849. The replacement removes the process-wide SSL injection while preserving Telegram, Discord, and generic webhook verification with outbound-only native TLS contexts backed by the OS CA store plus certifi. It also adds direct/proxied aiogram regression coverage.

PR #724, opened 2026-07-26
perf(crud): eliminate N+1 queries by eagerly loading relations

x0sina:
- Add selectinload for Admin.users, Admin.usage_logs, and Admin.role in get_admins to prevent sequential queries when not in compact mode
- Add selectinload for Group.users and Group.inbounds in get_group to eagerly load relations
- Add selectinload for Node.usage_logs in get_nodes, get_limited_nodes, and get_nodes_to_reset_usage to avoid loading attributes sequentially
- Replace individual db.refresh() calls with single batch re-fetch queries in bulk_reset_node_usage and update_users_proxy_settings
- Use .unique() on query results when eager loading relations to handle duplicates from joins
- Remove sequential load_admin_attrs, load_group_attrs, and load_node_attrs calls where relations are now eagerly loaded
- Use selectinload in _build_user_select_stmt to eagerly load all user relations (admin, next_plan, usage_logs, groups)
- Improves query performance by reducing database round trips and eliminating N+1 query patterns across admin, group, node, and user CRUD operations

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [ ] Tests / CI

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved data loading across administrators, groups, nodes, and users by using eager-loading and reducing per-record database work.
  * Made bulk proxy-setting updates and node usage resets more efficient by re-fetching updated records in batches.
  * Reduced repeated database dialect lookups during usage recording.
* **Bug Fixes**
  * Admin usage-limit warning reminders are now deduplicated and idempotent, with correct behavior when sending notifications fails for some administrators.
* **Database**
  * Enforced uniqueness for admin notification reminders using a dedicated constraint and normalized `threshold` values (including non-null defaults with SQLite compatibility).
* **Tests**
  * Added unit tests for reminder idempotency and notification failure handling.

PR #851, opened 2026-09-03
perf(nodes): avoid materializing reset history in lists

dr-hoseyn:
## Summary

The dashboard node list is polled every 10 seconds. It previously `selectinload`ed every `NodeUsageResetLogs` row for the page and materialized those rows as ORM objects just to calculate `lifetime_uplink` and `lifetime_downlink`.

This change:

- calculates reset uplink/downlink with indexed correlated `SUM` expressions in the node query
- keeps reset-history rows unloaded for the dashboard list
- preserves the existing default loading behavior for callers that actually need the history objects
- adds regression coverage for aggregate values, empty history, and the operation wiring
- adds a reproducible before/after benchmark

## Not a duplicate

I checked all local/remote branches and open PRs for node list/reset-history aggregation. No existing change implements this. In particular, #615 targets pools, indexes, usage writes, and health checks; #724 fixed N+1 loading by eagerly loading relations but still materializes the complete node reset history.

## Benchmark

Command: `python -m scripts.benchmark_node_list_usage`

SQLite in-memory, matching the dashboard page size of 15 nodes, with 1,000 reset records per node; median of 5 timed runs:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Median query time | 208.33 ms | 7.49 ms | -96.40% |
| Peak traced memory | 25.39 MiB | 0.13 MiB | -99.49% |
| SELECTs | 3 | 2 | -1 |
| Reset-log ORM objects | 15,000 | 0 | -100% |
| Session identity map | 15,015 | 15 | -99.90% |

The benchmark also asserts that lifetime usage is identical in both paths.

## Verification

- focused node aggregate/loading tests: 5 passed
- non-API suite excluding the pre-existing API-coupled review test: 166 passed, 2 skipped
- Ruff check and format check: passed
- `git diff --check`: passed

The local API test harness remains blocked by its existing database setup: the on-disk test DB references missing Alembic revision `a4d8c7e91b32`, while the in-memory override creates tables on a connection different from the app session (`no such table: jwt`).

## Summary by CodeRabbit

- **New Features**
  - Node lists now display accurate lifetime uplink and downlink usage totals, including usage accumulated across reset periods.

- **Performance Improvements**
  - Node usage totals load more efficiently without loading the full usage-log history, reducing database and memory overhead for large datasets.

- **Bug Fixes**
  - Corrected lifetime usage reporting for nodes with and without recorded usage history.


dr-hoseyn:
@coderabbitai review

PR #853, opened 2026-09-03
perf(nats): avoid repeated serialization when sizing user batches

dr-hoseyn:
## Summary

Bulk user updates dispatched through NATS repeatedly JSON-encode the growing batch just to decide whether the next user fits. With the default 100-user batch size, sizing 1,000 users serializes 50,500 user objects and generates about 32 MB of temporary encoded data before publication.

Measure the empty command envelope once and each user once with the existing wire encoder, then accumulate exact byte lengths, including array commas. Append users without copying each partial batch. This reduces synchronous CPU work during bulk node synchronization while preserving batch boundaries, order, payload limits, and the wire format.

### Before / after

Local measurements against `dev` at `234ab68c`, using synthetic dictionaries with a 600-byte payload field, batch size 100, and a 900,000-byte command limit:

| Metric | Before | After |
| --- | ---: | ---: |
| Sizing time, 1,000 users (median) | 97.537 ms | 7.419 ms |
| User objects encoded during sizing, 1,000 users | 50,500 | 1,000 |
| Temporary bytes encoded during sizing, 1,000 users | 32,057,995 | 680,938 |
| Output for 1,000 users | 10 commands / 634,360 bytes | Identical |
| Sizing time, 10,000 users (median) | 930.614 ms | 64.365 ms |
| User objects encoded during sizing, 10,000 users | 505,000 | 10,000 |
| Temporary bytes encoded during sizing, 10,000 users | 321,088,495 | 6,818,938 |
| Output for 10,000 users | 100 commands / 6,353,590 bytes | Identical |

The 10,000-user sizing measurement is approximately 14.5 times faster. Total encoder calls are N+1 after this change (including one empty envelope), compared with N in these baseline fixtures; the improvement comes from encoding much less data per call. Temporary byte totals are cumulative encoder output, not peak memory or network traffic.

Timing method: Python 3.14.6 on Windows 11; 3 warmups and 11 samples per implementation, alternating measurement order, in one process. The temporary benchmark harness loaded the baseline helper from Git and the updated helper from the working tree, used the actual `encode_node_command`, and asserted identical encoded commands. Counters were collected separately from timing. Data preparation, final publication encoding, and network/server work were excluded. These are local measurements of the sizing helper, not end-to-end production latency claims.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (No public interface changes.)

## Testing

```text
python -m pytest tests/test_node_sync.py tests/test_node_manager_sync.py tests/test_node_manager.py -q
26 passed

python -m ruff check --no-fix app/node/sync.py tests/test_node_sync.py
All checks passed!

python -m ruff format --check app/node/sync.py tests/test_node_sync.py
2 files already formatted

git diff --check
Passed
```

The expanded batching tests cover exact byte limits and one byte either side, nested JSON with Persian text/emoji/escaped characters, count limits and their minimum clamp, the 1,024-byte payload minimum, oversized singleton warnings, empty input, unchanged objects/order, and linear serialization work. Before the implementation change, the 16 behavior cases passed and the new performance regression failed with `50500 != 1000`; all 17 batching cases now pass.

The deterministic performance regression can be reproduced independently of timing with:

```text
python -m pytest tests/test_node_sync.py::test_node_update_users_nats_chunk_sizing_encodes_each_user_once -q
```

## Screenshots

Not applicable; backend-only change.

## Notes for reviewers

Only the NATS batch-sizing helper and its tests change. The size calculation uses the existing compact JSON encoder and accounts for the single comma between array entries. It does not change publication or serialization of final commands. A single user that exceeds the configured limit still forms its own batch and logs a warning, preserving existing behavior. No database migrations, new dependencies, or live NATS service are required for the tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved user synchronization batching to handle payload limits and batch sizes consistently.
  * Preserved oversized individual users while issuing an appropriate warning.
  * Avoided unnecessary repeated serialization during synchronization, improving efficiency.

* **Tests**
  * Added coverage for payload boundaries, batch limits, empty inputs, oversized users, warnings, and serialization behavior.


dr-hoseyn:
@coderabbitai review

PR #850, opened 2026-09-03
perf(users): avoid materializing reset history in lists

dr-hoseyn:
## Summary

- add opt-in aggregate loading for lifetime user traffic in the list query
- make the dashboard user list calculate reset usage in SQL instead of materializing every `user_usage_logs` row
- preserve the existing eager-loading default for non-dashboard callers
- add regression coverage and a reproducible before/after benchmark

## Benchmark

Command:

```console
python -m scripts.benchmark_user_list_usage
```

Environment: Windows, Python 3.14, in-memory SQLite; 50 users with 1,000 reset-history rows each; median of 5 warmed runs. Peak memory is measured in a separate `tracemalloc` run.

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Median query + serialization time | 1040.94 ms | 15.25 ms | -98.54% |
| Peak Python allocations | 73.64 MiB | 0.29 MiB | -99.61% |
| SELECT statements | 5 | 4 | -20% |
| Usage-log ORM objects | 50,000 | 0 | -100% |
| Session identity-map objects | 50,050 | 50 | -99.90% |

The response still returns the exact same `lifetime_used_traffic` values, including users with no reset history.

## Tests

- `python -m pytest tests/test_user_list_usage_aggregate.py -q` (2 passed)
- `python -m pytest tests --ignore=tests/api --ignore=tests/test_review_admins_unit.py -q` (165 passed, 2 skipped)
- `ruff check app/db/crud/user.py app/operation/user.py tests/test_user_list_usage_aggregate.py scripts/benchmark_user_list_usage.py`
- `ruff format --check app/db/crud/user.py app/operation/user.py tests/test_user_list_usage_aggregate.py scripts/benchmark_user_list_usage.py`

## Summary by CodeRabbit

* **Performance**
  * Improved user-list loading by displaying lifetime traffic totals without loading individual usage history records.
  * Added configurable loading options for usage logs and lifetime usage data.

* **Bug Fixes**
  * User lists now correctly show aggregated lifetime traffic while keeping detailed usage history unloaded.

* **Tests**
  * Added regression coverage for lifetime traffic aggregation and efficient user-list loading.

PR #854, opened 2026-09-03
perf(bulk): batch user reloads and aggregate reset usage

dr-hoseyn:
## Summary

Bulk expiry, data-limit and group changes loaded admin/role, next plan, reset history and groups separately for each returned user. Bulk proxy updates already batched those relationships, but still materialized every reset-history row. For 100 affected users with 1,000 reset records each, this loaded 100,000 history objects after a single bulk operation.

Reuse the existing user-select builder to join scalar relationships, batch-load admin roles/groups and calculate lifetime usage with an indexed SQL aggregate. Apply the same reload to all five bulk operations. Group removal and proxy updates capture IDs before mutation rather than loading full user objects twice.

The reload uses `populate_existing=True`, which also fixes stale group collections: with `expire_on_commit=False`, a user whose groups were already loaded could retain the old membership after a bulk add/remove. The returned object now reflects the committed membership.

## Before / after

Local benchmark: **100 users, 1,000 reset records per user**, one shared admin/role and one group. Complete CRUD call, including target selection, mutation, commit and reload:

| Operation | Median before → after | SQL statements before → after | Peak Python allocations before → after |
| --- | ---: | ---: | ---: |
| Data limit | 1,370.881 → 23.604 ms | 306 → 6 | 115.061 → 0.777 MiB |
| Expiration | 1,264.995 → 29.114 ms | 306 → 6 | 115.061 → 0.774 MiB |
| Add groups | 1,416.775 → 25.759 ms | 306 → 6 | 115.095 → 0.800 MiB |
| Remove groups | 1,385.899 → 24.804 ms | 305 → 6 | 115.060 → 0.746 MiB |
| Proxy settings | 1,958.861 → 23.752 ms | 6 → 5 | 152.659 → 0.759 MiB |

**Reset-history ORM objects: 100,000 → 0 in every case.** The benchmark asserts equal affected counts and complete serialized `UserResponse` values for both implementations, including lifetime usage, updated fields and group membership.

Method: Python 3.14.6 on Windows, SQLite in memory, baseline `234ab68c`. One warmup per version, five timing samples per version with alternating execution order, fresh sessions and equivalent fixture state for each sample. SQL/object counts and `tracemalloc` peaks are measured in separate untimed runs. Fixture setup/reset, response validation and node dispatch are excluded. Peak allocations are not process RSS or database memory. These are local CRUD measurements, not production PostgreSQL/MySQL or end-to-end API latency claims.

Reproduce from the repository root:

```sh
python scripts/benchmark_bulk_user_loading.py --baseline 234ab68c --users 100 --logs 1000 --runs 5
```

The script uses its own in-memory database. A separate `--users 2 --logs 0 --runs 1` smoke run also verified equal responses with no history.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Testing

**46 passed**, including 20 new regression cases and existing node-sync, WireGuard and admin tests:

```powershell
$env:DATABASE_URL = 'sqlite+aiosqlite:///:memory:'
.venv/Scripts/python.exe -m pytest tests/test_bulk_user_loading.py tests/test_node_sync.py tests/test_node_manager_sync.py tests/test_wireguard_alloc_unit.py tests/test_review_admins_unit.py -q
```

The database override isolates the existing admin test suite from a local `test.db` referencing a migration absent from this branch. The new bulk tests always create their own in-memory database.

Regression coverage includes all five operations, full response validation, absence of history-object loads, positive/negative adjustments, unchanged and empty scopes, cached membership refresh, and query growth across SQLAlchemy's 500-parent select-in batch boundary. Before the fix, the initial regressions reproduced history loads, stale memberships, and **192 SQL statements for 62 returned users**; the latter now uses 6.

`ruff check --no-fix`, `ruff format --check` on all three changed files, and `git diff --check` passed.

## Checklist

- [x] I updated documentation, translations, or examples if needed. Reproduction instructions are included in the benchmark.
- [x] I checked database migrations when models or schema changed. No model/schema changes.

## Screenshots

Not applicable; no UI changes.

## Notes for reviewers

- Independent of #850/#851/#852/#853 and #827; based directly on `dev`. Only `app/db/crud/bulk.py` changes application behavior.
- Existing filtering, status transitions, affected counts and sync dispatch remain in place. Admin roles and next plans remain available, and proxy results preserve captured ID order. Direct `usage_logs` access is no longer eagerly supported by these bulk results; current callers only use them for allocation/node sync. Lifetime values remain available through the aggregate.
- SQL still scans the matching indexed reset rows for the sum; this removes transfer/materialization of history, not the database's aggregation work. Relationship queries grow by select-in batches for larger populations, not one fixed query count for every size. Small populations with no history can have slightly more query bookkeeping/allocation overhead.
- No dependencies, migrations or changes to retention. Reset-history rows remain intact.

## Summary by CodeRabbit

- **Improvements**
  - Bulk user-management operations now more reliably reflect updated group memberships, expiration status, data limits, and proxy settings.
  - Results remain accurate when processing large user selections and users with previously loaded information.
  - No-op and empty selections are handled cleanly, and only newly affected users are reported for certain status changes.
- **Performance**
  - Bulk operations are more efficient when processing many users.


dr-hoseyn:
@coderabbitai review

PR #855, opened 2026-09-03
perf(admin): avoid history loads and per-user inbound queries

dr-hoseyn:
## Summary

Changing an admin's blocking status or activating/disabling their users currently loads every selected user's reset history and next plan. Restoring active users also performs a separate inbound query per user. An admin with 100 users and 100,000 reset records therefore materializes 100,000 unnecessary history objects before dispatch.

Add `get_admin_users_for_node_sync`, scoped by admin ID and optional statuses. It selects the four User columns needed by node synchronization, joins the admin and role, and batches groups and their inbounds. Use it for block/unblock transitions and the two admin-wide activation/disable operations. This avoids history/next-plan loading and per-user inbound queries while preserving user selection, blocking rules, status changes, and node payload content.

## Type of change

- [x] Refactor / cleanup
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. No documentation changes were needed.
- [x] I checked database migrations when models or schema changed. No schema changes.

## Testing

**Before/after benchmark:** local Windows, Python 3.14.6, in-memory SQLite, 100 active users under one admin, one shared group/inbound, 1,000 reset records per user. Baseline operation and CRUD loader were executed from `234ab68c`. One warmup per version was followed by five alternating samples; the table shows medians. SQL/history counters and `tracemalloc` peaks came from separate instrumented runs, not the timing runs.

| Operation | Before ms | After ms | SQL before -> after | History objects before -> after | Python peak MiB before -> after |
| --- | ---: | ---: | ---: | ---: | ---: |
| Block admin's users | 2145.658 | 10.706 | 5 -> 3 | 100,000 -> 0 | 192.720 -> 0.566 |
| Restore admin's users | 2050.641 | 11.080 | 106 -> 3 | 100,000 -> 0 | 192.719 -> 0.562 |

The measured scope was user loading and protobuf preparation through captured dispatch. Fixture creation, loading/changing the admin, database/session setup, message comparison, network delivery, and node processing were excluded. The one-off benchmark compared every complete serialized message between versions, normalizing user and inbound order, and verified counts and expected inbound assignments. It used an isolated in-memory database and captured dispatch, so no configured database or node was contacted.

**44 tests passed:** 18 new regression cases plus the existing node-sync, node-manager, WireGuard allocation, and admin-review suites. New coverage exercises actual operations and serializers with only dispatch captured: admin scope, limited/disabled role policies, block/unblock and no-op transitions, active/on-hold/disabled/limited/expired users, enabled/disabled groups, duplicate inbound tags, empty groups/admins/status filters, detached serialization, preserved proxy credentials and status mutations, and zero history/next-plan materialization. Query bounds cover 62 users (at most 3 queries) and 512 users (at most 5 queries).

Before the fix, the initial 14-case regression set had 9 failures on unwanted history loads and 5 passes; all 18 final cases pass after the fix.

```powershell
$env:DATABASE_URL = 'sqlite+aiosqlite:///:memory:'
.venv/Scripts/python.exe -m pytest tests/test_admin_user_sync_loading.py tests/test_node_sync.py tests/test_node_manager_sync.py tests/test_wireguard_alloc_unit.py tests/test_review_admins_unit.py -q --tb=short
```

Ruff lint, Ruff format checks, and `git diff --check` pass for all changed files. A two-user, zero-history benchmark smoke run also passed message equality.

## Screenshots

Not applicable.

## Notes for reviewers

- The new loader returns partially loaded ORM users specifically for node dispatch, not API response validation. The admin-wide deletion path retains the full loader because it builds notification/API models.
- Query counts grow with select-in batches and group/inbound cardinality; three queries is the measured 100-user fixture, not a universal constant. User/group/inbound memory still scales with the selected population; this is not streaming or bounded-memory dispatch.
- Small populations without reset history can allocate slightly more due to the eager inbound path (two-user smoke: approximately 0.109 -> 0.132 MiB for block). Production PostgreSQL/MySQL runtime and end-to-end network latency were not benchmarked; timings above are local measurements.
- Based directly on `dev`; no dependency on #850, #851, #852, #853, or #854. #827 optimizes the separate group-update synchronization path; this change targets admin transitions and admin-wide user actions.


dr-hoseyn:
@coderabbitai review

ImMohammad20000:
why the script exists?

dr-hoseyn:
The script has been removed. Thanks.

PR #856, opened 2026-09-03
perf(dashboard): defer user dialogs and chart dependencies

dr-hoseyn:
## Summary

The app shell preloaded Recharts, and entering the dashboard/users routes eagerly imported the user editor and usage/subscription dialogs even when closed. Load those three dialogs on first open and let Vite/Rolldown split their dependencies automatically, removing manual vendor groups that pulled chart code into the shell.

The shared deferred boundary shows a cancellable loading dialog while keeping the page mounted. After first open, it retains the existing dialog so its close effects, form state and exit animation still run. All user-editor call sites and the editor's nested usage dialog use the wrappers; existing chunk-load recovery is reused.

The generated service worker precaches the entry script and its static JavaScript dependencies, plus existing HTML/CSS/image/font assets. Deferred JavaScript uses a bounded CacheFirst runtime cache after it is requested. No automatic service-worker registration is introduced.

## Type of change

- [x] Refactor / cleanup
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed (reproduction below; existing loading translation).
- [x] I checked database migrations when models or schema changed (not applicable).
- [x] I did not include secrets, tokens, private keys, or unrelated changes (browser fixture credentials are synthetic).

### Before / after

Baseline: `234ab68cdf4ca01bef5f5bee6bddbe1068059fc8` on `dev`. Both production builds used the same installed dependencies (Node 24.15.0, Vite 8.2.1), separate fresh output directories, and `--manifest`. Measurements sum unique JavaScript files and their individually gzipped sizes; kB means 1,000 bytes.

| Dependency set | JS files before -> after | Raw kB before -> after | gzip kB before -> after | gzip reduction |
| --- | ---: | ---: | ---: | ---: |
| HTML app shell, before a route loads | 15 -> 14 | 1,239.346 -> 738.843 | 350.612 -> 219.849 | 37.3% |
| Dashboard static imports, before its deferred chart | 104 -> 70 | 2,283.727 -> 1,492.545 | 604.668 -> 399.543 | 33.9% |
| Users page, dialogs closed | 114 -> 106 | 2,437.704 -> 1,907.953 | 648.914 -> 524.400 | 19.2% |
| Dashboard including its deferred chart | 105 -> 76 | 2,293.546 -> 1,942.255 | 608.148 -> 522.708 | 14.0% |

These are dependency-set sizes, not measured page-load times. The dashboard still displays its chart after the existing 250 ms delay. Opening usage downloads its chart code, and first-opening a deferred dialog requires a chunk fetch. CSS, fonts, translations, API responses, service-worker precaching, HTTP headers and browser cache effects are excluded.

### Service-worker precache and dashboard close behavior

Compared with the original PR head `a82315588650d0f7625f244c0ff14564673a87a4`, using fresh builds with the same dependencies:

| Generated precache | Original PR | Updated PR |
| --- | ---: | ---: |
| JavaScript files | 292 | 14 |
| Total entries, including non-JS assets | 317 | 39 |
| Total precache size reported by Workbox | 10,122.05 KiB | 2,348.57 KiB |

This is a **76.8% reduction in generated precache bytes**, not a page-load-time measurement. The 14 retained JavaScript files are exactly the entry's transitive static imports. Deferred route, dialog, chart and editor JavaScript is excluded; successful same-origin scripts under `statics` are cached on demand for up to 30 days / 128 entries. Existing non-JS precaching remains. Offline availability of unvisited routes, translations or backend data is not promised.

The dashboard previously removed its user-dialog wrapper immediately on close, bypassing its retention behavior. It now always renders the lightweight wrapper, which loads the implementation only on first open and lets close effects/exit transitions run. The browser regression fails against the original PR head (`Dashboard unmounted the dialog before its close transition`) and passes after the fix, including reopening.

The build check validates the generated service worker's complete JavaScript precache against the shell dependency graph. The browser test explicitly installs the worker, verifies deferred dialogs/chart are absent from the precache, and checks that requested lazy JavaScript is runtime-cached and can load in a fresh document offline. It also confirms the precached HTML and entry script remain available offline. Registration is explicit only in the test because the app does not auto-register a worker.

### Reproduction

From `dashboard`, build the baseline and this branch into different, initially absent directories:

```sh
node node_modules/vite/bin/vite.js build --outDir <fresh-before-directory> --manifest
node node_modules/vite/bin/vite.js build --outDir <fresh-after-directory> --manifest
```

Use this branch's measurement/check script against both outputs:

```sh
node scripts/check-deferred-chunks.mjs <before-directory> --report-only
node scripts/check-deferred-chunks.mjs <after-directory>
```

Boundary assertions fail on the baseline and pass after the change: the three modal implementations must be dynamic entries, the shell/dashboard/users static dependency sets must exclude the chart renderer, and editing a user must not pull in the usage chart.

With Playwright available on Node's module search path:

```sh
node scripts/check-deferred-dialogs.cjs <after-directory>
```

Set `PLAYWRIGHT_CHANNEL=msedge` or `chrome` to use an installed browser, or use Playwright's Chromium. The test starts a localhost static server and mocks every API/external request with synthetic data. No backend/account is required. Verified locally with headless Edge:

- Closed users page requests none of the three dialog chunks or the chart renderer.
- A held first-open chunk shows a loading dialog without hiding the users page; cancelling stays closed after the chunk arrives.
- Create/edit receive expected form values, without loading Recharts.
- Usage renders a real chart; subscription renders a QR canvas; closing/reopening both does not download the implementation chunk again.
- Narrow viewport (390 x 844) create/cancel, dashboard chart and login routes pass with no browser runtime errors.

Production build and new-file Prettier checks pass. `tsc -b --pretty false` reports **the same 26 diagnostics before and after, with identical output**; this PR does not fix or suppress them; the separate [PR #857](https://github.com/PasarGuard/panel/pull/857) addresses them. Breakdown: 15 subscription typing errors, 6 notification typing errors, 3 host typing errors, 1 user-filter callback mismatch, and 1 unused chart import. ESLint could not start because the existing configuration imports an unavailable `@eslint/js` package in this checkout.

## Screenshots

Not applicable to existing form/chart layouts; this is a loading-boundary change. The pending state uses existing Dialog and Spinner components. Desktop/mobile interactions were checked in the browser as listed above.

## Notes for reviewers

- Based directly on `dev`; no dependency on other open performance PRs.
- Automatic chunking changes asset names/dependency groups across the dashboard. Existing React deduplication and chunk-load recovery remain. [Rolldown documents automatic splitting as its default](https://rolldown.rs/reference/OutputOptions.codeSplitting).
- Dialogs stay mounted after first open for their parent's lifetime, preserving existing closed-state behavior. This does not address the separate action-button store retention issue.
- Build-directory cleanup, polling, Node/Host dialogs and editor-language bundles are outside this change. No dependency, API or database schema changes.

## Summary by CodeRabbit

* **Performance**
  * Improved dashboard load performance by loading user, usage, subscription, and chart-related dialogs only when needed.
  * Added loading feedback while deferred dialogs open.
  * Reduced unnecessary downloads when reopening dialogs.

* **Bug Fixes**
  * Improved reliability when deferred dialog chunks fail or loading is cancelled.
  * Preserved dialog state and form values across deferred editor loading.
  * Improved offline availability for core dashboard content and previously loaded deferred dialogs.
  * Verified deferred dialogs across dashboard, login, and mobile views.


dr-hoseyn:
@coderabbitai review

dr-hoseyn:
@coderabbitai Full review

PR #859, opened 2026-09-03
fix(hwid): serialize device registration across workers

dr-hoseyn:
## Summary

Concurrent subscription requests with different HWIDs can both pass the device-count check and consume the same final slot. Move the check into registration and serialize it with a database lock, so only one new device is admitted while repeat requests from an existing device remain allowed.

PostgreSQL/MySQL lock the parent user row; SQLite acquires a write lock before reading the HWID set. End the validation read transaction first so MySQL's default REPEATABLE READ cannot reuse an older snapshot after waiting for the lock.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Not needed.)

## Testing

- Reproduced the bug on the unchanged backend: both different-device requests returned 200 when only one slot remained, for limits of 1 and 3.
- `python -m pytest tests/api/test_hwid.py tests/api/test_user.py -k hwid -q --tb=short`: **13 passed, 83 deselected** on SQLite.
- The concurrency tests use separate database connections and pre-registration reads, and cover both different HWIDs competing for the last slot and duplicate requests for one HWID.
- Existing-device refresh at capacity still preserves its metadata and updates `last_used_at`; rejected registrations release the lock.
- Ruff check/format and `git diff --check` passed.
- Local database validation used SQLite. The new tests also run in the existing PostgreSQL, TimescaleDB, MySQL, and MariaDB CI jobs.

## Screenshots

Not applicable.

## Notes for reviewers

The registration helper still owns its commit and now returns whether the device was accepted. The operation maps rejection to the existing 403 response. Existing HWID policy resolution, unlimited registration, and subscription response behavior are unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved device registration to enforce account HWID limits reliably, including during simultaneous registration attempts.
  - Prevented users from exceeding their configured device limit.
  - Re-registering an existing device continues to work without consuming an additional slot.
  - Registration requests that exceed the device limit now return a clear 403 error.

### DIFF
--- a/app/db/crud/hwid.py
+++ b/app/db/crud/hwid.py
@@ -7,7 +7,7 @@ from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.sql.schema import Table
 
-from app.db.models import UserHWID
+from app.db.models import User, UserHWID
 
 
 async def get_user_hwids(db: AsyncSession, user_id: int) -> list[UserHWID]:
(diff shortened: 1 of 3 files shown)


dr-hoseyn:
@coderabbitai review

PR #860, opened 2026-09-03
perf(nodes): avoid loading reset history when selecting due nodes

dr-hoseyn:
## Summary

The periodic node reset job loads every reset-history row for all candidate nodes, even though its SQL query already calculates the latest reset timestamp. Select that timestamp alongside each node and reuse it for the existing schedule checks. Candidate selection now executes one query and materializes no reset-log objects; daily, weekly, monthly, yearly, and interval reset rules are unchanged.

Refresh the post-reset nodes with `populate_existing=True` so notification consumers receive current history and consistent database-normalized timestamps. Without this, SQLite can mix the newly inserted timezone-aware timestamp with older timezone-naive timestamps and fail when selecting the latest log.

## Type of change

- [x] Bug fix
- [x] Refactor / cleanup
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Not needed.)

## Testing

- `python -m pytest tests/test_node_reset_candidates.py tests/test_node_usage_logs_load.py -q` — 21 passed locally on SQLite.
- The 18 new cases fail on the original implementation's extra query/history loading, and pass after the change.
- Coverage includes all four calendar schedules, due/not-due nodes, latest timestamp independent of insertion order, creation-time fallback, interval/status filtering, and bulk reset followed by response serialization and latest-log lookup.
- Ruff check/format and `git diff --check` passed.

## Screenshots

Not applicable.

## Notes for reviewers

Checked current `dev` and open PR file coverage. #851 changes node list aggregation but does not modify `get_nodes_to_reset_usage`; this is an independent change targeting `dev`.

Full history still loads after an actual reset for the existing response/notification consumers. The optimization removes history loading from candidate selection, including nodes whose scheduled reset is not yet due.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node usage reset processing to consistently use the latest reset timestamp.
  * Ensured refreshed usage data and generated timestamps are shown correctly after bulk resets.
  * Confirmed that reset operations clear usage, restore eligible status, and record usage logs accurately.

* **Tests**
  * Added coverage for scheduled and interval-based reset behavior, including daily, weekly, monthly, and yearly schedules.
  * Added validation for disabled nodes and nodes configured to skip resets.


dr-hoseyn:
@coderabbitai review

PR #801, opened 2026-08-17
feat(cleanup): add automatic retention policies

dr-hoseyn:
## Summary

- add independent automatic retention settings for expired accounts, user/node traffic history, and node system metrics
- delete historical rows hourly in bounded batches with per-run safety limits and time-based indexes
- preserve the legacy `USERS_AUTODELETE_DAYS` behavior during migration, including immediate deletion with `0`
- add a responsive Cleanup settings card with validation, disabled/unlimited states, RTL support, and four locales

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I tested the change locally or explained why it cannot be tested.
- [x] I added or updated tests for behavior changes.
- [x] I updated documentation, translations, or examples if needed.
- [x] I checked database migrations when models or schema changed.
- [x] I did not include secrets, tokens, private keys, or unrelated changes.

## Testing

- `.venv/Scripts/python.exe -m ruff check .` — passed
- fresh SQLite `alembic upgrade head` with `USERS_AUTODELETE_DAYS=0` — passed; migrated value remained `0`
- `.venv/Scripts/python.exe -m pytest tests/api/test_settings.py tests/test_cleanup_retention.py -q` — 10 passed
- Prettier check for the new component and generated API types — passed
- Vite production build — passed
- browser validation at desktop and 390×844 mobile widths, in English LTR and Persian RTL — no horizontal overflow

### Cleanup retention settings — desktop, dark mode

Cleanup automatic retention settings

Also verified at 390×844 mobile width and in Persian RTL with no horizontal overflow.

## Notes for reviewers

- Defaults are unlimited for expired accounts, 90 days for traffic history, and 30 days for node system metrics.
- A per-account `auto_delete_in_days` override still takes precedence over the global expired-account policy.
- Retention deletes use 10,000-row transactions and stop at 1,000,000 rows per table per run; remaining rows are retried on the next hourly run.
- `None` means keep forever. The settings CRUD path now preserves explicit nested `None` values while continuing to ignore omitted top-level settings sections.
- The contribution guide references `next`, but that branch is not present on the upstream remote. Current project PRs target `dev`, so this PR does too.

## Summary by CodeRabbit

* **New Features**
  * Added configurable retention settings for expired accounts, usage history, and node statistics, including indefinite retention.
[...]

PR #789, opened 2026-08-14
perf(db): bound multi-worker connection pools

dr-hoseyn:
## Summary

- reduce the default non-SQLite pool ceiling from 85 to 10 connections per process so documented multi-worker deployments do not multiply unsafe defaults
- preserve explicit pool overrides, validate unsafe values, and expose `SQLALCHEMY_POOL_TIMEOUT`
- log the credential-free per-process and configured service connection ceilings at startup
- document the deployment-wide formula and cover PostgreSQL, MySQL, SQLite, override, validation, and summary behavior

Closes #788

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [x] Documentation
- [x] Tests / CI

## Checklist

- [x] I tested the change locally or explained why it cannot be tested.
- [x] I added or updated tests for behavior changes.
- [x] I updated documentation, translations, or examples if needed.
- [x] I checked database migrations when models or schema changed. (No schema change.)
- [x] I did not include secrets, tokens, private keys, or unrelated changes.

## Testing

- `.venv/Scripts/python.exe -m ruff check .` — passed
- focused pool/app-factory tests — 17 passed
- full SQLite suite on a clean database — 535 passed, 2 skipped

## Screenshots

Not applicable (backend/configuration-only change).

## Notes for reviewers

- `SQLALCHEMY_POOL_SIZE` and `SQLALCHEMY_MAX_OVERFLOW` remain per-process settings and explicit environment values retain their existing meaning.
- Defaults change from `25 + 60` to `5 + 5`, limiting four Uvicorn workers to 40 potential connections instead of 340. Operators with a deliberately sized external pool can keep larger values explicitly.
- SQLite continues to omit QueuePool options.
- Startup summaries never include the database URL, host, username, or password.
- No migration or API contract change.

## Summary by CodeRabbit

- **Improvements**
  - Database connection pooling now uses safer, bounded defaults to improve resource management.
  - Added configurable connection timeouts and validation for pool settings.
  - Startup logs now report effective database connection limits without exposing connection details.
  - SQLite connections avoid unsupported pooling options, while MySQL retains appropriate connection settings.

- **Bug Fixes**
  - Invalid pool configurations and process counts are now rejected early with clear validation errors.

PR #865, opened 2026-09-04
perf(statistics): reduce query load on large installs

### LINKED ISSUE #858: [Bug]: Opening the Statistics page saturates all CPU cores on large installs
### Before you submit

- [x] I am using the latest available PasarGuard version.

### PasarGuard version

v5.3.0 (commit `aebf7256`)

### Installation type

Docker

### What happened?

On an install with a high user count, opening the **Statistics** page saturates the whole host. All 4 cores hit ~100% and load average climbs to ~21 within seconds, and it stays there for as long as the page is open — the panel becomes unresponsive for everything else. Closing the tab fixes it.

Two things stand out in the `htop` capture below. About 20 mysqld threads sit in the running state at once, so many heavy queries overlap rather than one running at a time. And the panel process is pinned at 99% too, so this isn't only the database.

### Steps to reproduce

1. Use an install with a high user count and several weeks of accumulated usage history.
2. Open `/panel/#/statistics` as owner, with `master` selected.
3. All cores saturate within seconds and stay saturated until the tab is closed.

### Expected behavior

A read-only dashboard page should not be able to saturate the host, or degrade the panel for other admins and for subscription traffic.

### Logs and error messages

```shell
# Nothing is logged and nothing crashes — it just saturates. See the htop screenshot below.
```

### Environment

```
OS: Linux, 4 vCPU / 7.56 GB RAM
Database: MySQL (local)
```

### Screenshots

`htop` taken while the Statistics page was open:

Image

### Additional context — code that looks responsible

Three of the five requests this page fires on mount:

1. **`GET /api/users/counts/{metric}`** — `get_user_count_metric_stats` (`app/db/crud/user.py`) runs two queries for one chart: an ungrouped total plus the grouped series. Both are `COUNT(DISTINCT user_id)` over `node_user_usages` grouped by `DATE_FORMAT(CONVERT_TZ(created_at, ...))`, over a 7-day default range. That table gets a row per user per node per 10-minute bucket and has no retention job. `_build_user_count_query_parts` also joins `users` unconditionally, including for `metric=online`, which reads nothing from `users`.

2. **`GET /api/users/sub_update/chart`** — default range is `all` (start `2000-01-01`), also two queries. `user_subscription_updates` has no index starting with `created_at`, which is the only column this query filters on.

3. **`GET /api/system/users`** — polled every 2s from this page, a full pass over `users` each time.

None of them are cached, and there's no cap on the requested span or period.

The fix is yours to choose — there's related work in flight already (#786 / #787, #801, #604). Happy to test a patch or send `EXPLAIN` output from this install.

dr-hoseyn:
## Summary

- reduce `/api/users/sub_update/chart` from two history aggregations to one by deriving pie totals from the period rows
- remove unnecessary `users` joins from owner-wide subscription history and unscoped online-count queries
- add a `created_at` index for owner-wide `user_subscription_updates` time ranges
- default subscription history to one month and move aggregate user-stat polling from 2s to 30s while keeping live resource polling at 2s
- document the query-load safeguards, migration caveat, and directional benchmark

Closes #858

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [x] Documentation
- [x] Tests / CI

## Checklist

## Testing

- `.venv/Scripts/python.exe -m pytest -q tests/test_statistics_query_load.py tests/test_system_user_metrics.py tests/api/test_usage_functions_timezone.py` — 32 passed
- targeted Ruff lint and format checks — passed
- `npm run build` in `dashboard` — passed
- fresh SQLite `alembic upgrade head`, downgrade to `7c4bd5128e62`, and upgrade again — passed; the new index appeared, disappeared, and reappeared as expected
- MySQL DDL compilation — `CREATE INDEX idx_user_subscription_updates_created_at ON user_subscription_updates (created_at)`

Directional in-memory SQLite benchmark, 400,000 rows per history table, median of seven warmed runs:

| Query shape | Before | After | Change |
| --- | ---: | ---: | ---: |
| Unscoped online count | 406.16 ms | 325.05 ms | -20.0% |
| Subscription chart | 896.68 ms | 502.13 ms | -44.0% |

These numbers validate direction and query shape; they are not a production MySQL latency guarantee. The reporter's offered `EXPLAIN ANALYZE` output remains the best follow-up validation on the affected installation.

The three standalone `tests/api/test_user.py` permission cases were also attempted, but the current API harness did not seed the `testadmin` login and failed in the existing `access_token` fixture before reaching the endpoint. The operation path and response equivalence are covered by the new database-backed test.

## Screenshots

Not applicable. The UI changes only adjust polling and the initial time range; layout is unchanged.

## Notes for reviewers

I checked related prior and in-flight work before implementing this:

- #791 is already merged and reduces `/api/system/users` to one SQL statement; this PR keeps it and reduces how often the full aggregate runs on the Statistics page.
- #604 slows polling only when the document is hidden. This issue reproduces while the tab is visible, so the changes are complementary.
- #801 adds retention policies, which limit future table growth but do not remove the current duplicate scan or joins.
- #789 bounds connection pools and #852 forwards request cancellation; neither changes these query shapes.
- #787 targets PostgreSQL write amplification and intentionally leaves MySQL unchanged.

The new index must scan the existing subscription-update table. Large MySQL/MariaDB installations should apply this migration during a maintenance window because DDL locking depends on server version and table configuration. All-time history remains available as an explicit chart selection.

## Summary by CodeRabbit

* **Performance**
  * Improved Statistics page query performance for subscription history and user metrics.
  * Added faster time-based processing for subscription update records.
  * Reduced unnecessary database lookups in common statistics views.

* **Usability**
  * Subscription charts now default to displaying the past month.
  * Statistics refresh less frequently to reduce system load.

* **Documentation**
  * Added guidance on Statistics page performance behavior and verification.

* **Tests**
  * Added coverage for subscription chart totals and query behavior.


dr-hoseyn:
@coderabbitai review

ImMohammad20000:
remove docs

dr-hoseyn:
Done

PR #866, opened 2026-09-04
fix(node): recover from slow Start timeouts

### LINKED ISSUE #861: [Bug]: Node stuck in Stop/Start loop when Start exceeds the 60s timeout cap
### Before you submit

### PasarGuard version

v5.3.0

### Installation type

Docker Compose

### What happened?

On a WireGuard core with a high peer count, the node's `Start` RPC takes longer
than 60 s to return (61 s here). Start is called without an explicit timeout, so
it falls back to `default_timeout`, which is capped at 60 both in
`app/models/node.py` (`ge=3, le=60`) and in the dashboard form — so no allowed
value lets it finish.

After the timeout the panel cannot recover: the lifecycle lease is released with
`observed = BROKEN`, `node_health_check` reconnects every 10 s (its in-progress
guard only covers `observed is HEALTHY`), and because the state is `BROKEN` the
panel refuses to attach to the node that is in fact already running — it issues
Stop and a full Start instead. The interface is rebuilt about every 80 s and its
peers stay offline, while the node logs the Start as successful.

### Steps to reproduce

1. Add a WireGuard core on a node with enough peers that the node's `Start` RPC
   takes longer than 60 s to return.
2. Let the panel connect the node, or call `POST /api/node/{id}/reconnect`.
3. Watch the node status in the panel and `nodes.status` / `nodes.message` in
   the database.
4. In Modify Node, try to raise "Default Timeout" above 60.

### Expected behavior

`default_timeout` should allow a value large enough for a big core to finish
starting, and a panel-side timeout should not leave a running node in a state
that gets Stop plus a full Start every ~80 s.

### Logs and error messages

```shell
Node (panel IP redacted):

    2026/09/01 06:48:50 Starting Node: v0.5.4
    2026/09/01 06:48:51 config loaded in 4.8e-07 second.
    ... peer collection, all within the same log second ...
    2026/09/01 06:49:52 wireguard host routing: wg interface "pm-wg-de", NAT egress "eth0" (masquerade, egress_only=true, nat_disabled=false)
    2026/09/01 06:49:52 wireguard started, Version: v1.0.20260223
    2026/09/01 06:49:52 IP: <panel>, Method: Start, Status: Success
    2026/09/01 06:50:09 wireguard shutdown complete
    2026/09/01 06:50:09 IP: <panel>, Method: Stop, Status: Success
    2026/09/01 06:50:10 IP: <panel>, Method: GetBaseInfo, Status: Success
    2026/09/01 06:50:10 config loaded in 5.21e-07 second.

Panel, `nodes` table for that node:

    status  = error
    message = Request timed out

and after a health tick lands during an in-flight Start:

    message = Node lifecycle operation already in progress for 19

Browser form validation blocks the workaround:

    Default timeout must be 60 or lower
```

### Environment

OS: Ubuntu 26.04
Database: MySQL 9.7.2
Python version: 3.14 (official panel Docker image)
Node.js version: N/A (prebuilt image)

### Relevant configuration

```yaml
Node record (panel side): connection_type=grpc, keep_alive=60,
default_timeout=60 (schema maximum), internal_timeout=15

Core: WireGuard, kernel interface, single /16 client subnet
Node: pasarguard/node v0.5.4, Docker, network_mode: host, cap_add: NET_ADMIN
```

### Screenshots or videos

Image
Image

### Additional context

Node-side the 61 s is not config parsing or peer collection — those finish
within the same log second — it sits between peer collection and the
host-routing step. Raising Default Timeout from 15 to 60 kept this working until
the peer count grew; 60 is now the ceiling.

dr-hoseyn:
## Summary

- Remove the artificial 60-second upper bound from node `default_timeout` in both the API validation and dashboard form.
- Treat a timed-out `Start` RPC as an ambiguous outcome: probe `GetBaseInfo` and attach if the remote core finished starting.
- Let the health checker recover lifecycle state with `observed=BROKEN` and `desired=HEALTHY` before issuing a destructive reconnect.
- Require `GetBaseInfo.started` before attaching to avoid marking a stopped core healthy.
- Add regression tests for slow startup, late success, lifecycle reconciliation, and reconnect suppression.

Closes #861

## Why

The panel used `default_timeout` for `Start`, while both backend and frontend capped that value at 60 seconds. When a large WireGuard core finished just after that deadline, the bridge recorded the lifecycle as `BROKEN` even though the remote core was already running. The next health tick then replaced the controller, sent `Stop`, and started the core again, producing the reported loop.

A client-side timeout does not prove the remote operation failed. This change reconciles the actual remote state before returning an error or starting another lifecycle operation.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [x] Tests / CI

## Checklist

- [x] Documentation, translations, or examples are not needed for this change.
- [x] No database migration is needed; the persisted column remains an integer.

## Testing

- `python -m ruff check --no-fix app/models/node.py app/operation/node.py app/jobs/node_checker.py tests/test_node_start_timeout.py`
- `python -m pytest -q tests/test_node_start_timeout.py tests/test_connect_concurrency.py tests/test_node_manager.py tests/test_nats_node_memory.py` — **15 passed**
- `prettier --check dashboard/src/features/nodes/forms/node-form.ts`

The full frontend typecheck is currently blocked by unrelated errors already present on `dev`. The full API suite is also blocked in this local environment by the existing `access_token` fixture failing to authenticate `testadmin`.

## Screenshots

Not applicable.

## Notes for reviewers

The database schema is unchanged. Existing timeout values remain valid, while operators with large cores can now configure values above 60 seconds. The recovery path also prevents a late successful startup from being immediately followed by `Stop` and another full `Start`.

## Summary by CodeRabbit

- **New Features**
  - Node startup and connection handling now better recovers from ambiguous or timed-out start operations when the node is already healthy or running.
  - Health checks can confirm an active node connection before attempting a reconnect.

- **Bug Fixes**
  - Reduced unnecessary connection failures and reconnect attempts after remote node startup timeouts.
  - Improved handling of nodes with inconsistent lifecycle status.

- **Validation**
  - Increased the maximum allowed node default timeout beyond 60 seconds while retaining the minimum of 3 seconds.

### DIFF
--- a/app/jobs/node_checker.py
+++ b/app/jobs/node_checker.py
@@ -135,12 +135,13 @@ async def process_node_health_check(db_node: Node, node: PasarGuardNode):
             return
 
         # Prefer shared lifecycle state so multi-worker local NOT_CONNECTED does not thrash Start.
-        # Trust observed HEALTHY only if we can attach, or another worker still holds an active lease.
+        # A BROKEN observation with desired HEALTHY can be an ambiguous client-side Start
+        # timeout: the remote core may have completed startup after the panel gave up.
         shared_state = await node.get_lifecycle_state()
         if (
             health is Health.NOT_CONNECTED
             and shared_state is not None
-            and shared_state.observed is LifecycleStatus.HEALTHY
+            and (shared_state.observed is LifecycleStatus.HEALTHY or shared_state.desired is LifecycleStatus.HEALTHY)
         ):
             attached = await NodeOperation._attach_if_running(node, db_node.name)
             if attached is not None:
@@ -154,9 +155,10 @@ async def process_node_health_check(db_node: Node, node: PasarGuardNode):
                 )
                 return
 
-            # Stale HEALTHY (owner gone / core unreachable): fall through to reconnect.
+            # Stale/failed desired-healthy state: fall through to reconnect only after
+            # the attach probe and active-lease check have both failed.
             logger.debug(
-                "[%s] Shared lifecycle HEALTHY but attach failed and no active lease; reconnecting",
+                "[%s] Shared lifecycle desired HEALTHY but attach failed and no active lease; reconnecting",
                 db_node.name,
             )
 
--- a/app/models/node.py
+++ b/app/models/node.py
@@ -50,7 +50,7 @@ class Node(BaseModel):
     data_limit: int = Field(default=0)
     data_limit_reset_strategy: DataLimitResetStrategy = Field(default=DataLimitResetStrategy.no_reset)
     reset_time: int = Field(default=-1)
-    default_timeout: int = Field(default=10, ge=3, le=60)
+    default_timeout: int = Field(default=10, ge=3)
     internal_timeout: int = Field(default=15, ge=3, le=60)
     proxy_url: str | None = Field(default=None, max_length=256)
 
@@ -180,7 +180,7 @@ class NodeModify(NodeCreate):
     data_limit: int | None = None
     data_limit_reset_strategy: DataLimitResetStrategy | None = None
     reset_time: int | None = None
-    default_timeout: int | None = Field(default=None, ge=3, le=60)
+    default_timeout: int | None = Field(default=None, ge=3)
     internal_timeout: int | None = Field(default=None, ge=3, le=60)
 
     model_config = ConfigDict(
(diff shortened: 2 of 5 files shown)


dr-hoseyn:
@coderabbitai review

PR #879, opened 2026-09-08
fix(security): bind generated Xray client inbounds to loopback

Rerowros:
## Summary

The seeded Xray client template listens on `0.0.0.0` without authentication. Change its inbound listeners to `127.0.0.1` so importing the subscription does not expose the client's local proxy to its network.

A data migration updates system `xray_subscription` templates while preserving document formatting and text outside `inbounds`. Non-system and non-Xray templates are unchanged; downgrade deliberately does not restore public listeners. Administrators can still intentionally edit their client templates.

## Scope and compatibility

Standalone on current `dev`. This PR owns the loopback migration and its test; the duplicate is removed from #756. It requires no routing, security-stack or Node PR. The migration follows the current `8e2f1a9c4b70` revision.

## Validation

The current five-database CI matrix completes the migrations and loopback regression, then reports 568 passed, 2 skipped and the same unrelated FinalMask fixture failure on each database. That baseline test is fixed separately in #889, whose complete database matrix passes.

- `pytest -q tests/test_xray_loopback_migration.py`: 1 passed, covering system/non-system templates, inbound-only replacement and formatting preservation.
- Fresh SQLite `alembic upgrade head`: passed; `alembic heads` reports one head (`a8c2d491e705`).
- Scoped Ruff and `git diff --check`: passed.

This changes the default binding in existing system templates; maintainers can review that policy independently of the synchronization work.

### DIFF
--- a/app/db/migrations/versions/a8c2d491e705_bind_xray_client_inbounds_to_loopback.py
+++ b/app/db/migrations/versions/a8c2d491e705_bind_xray_client_inbounds_to_loopback.py
@@ -1,0 +1,89 @@
+"""Bind generated Xray client proxy inbounds to loopback.
+
+Revision ID: a8c2d491e705
+Revises: 8e2f1a9c4b70
+Create Date: 2026-08-09
+"""
+
+import re
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "a8c2d491e705"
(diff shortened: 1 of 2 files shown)


ImMohammad20000:
This can be change by admins and its not a big deal for final user

PR #891, opened 2026-09-11
perf(nats): reduce idle user-sync overhead

DrSaeedHub:
## Summary

Repeated queue discovery creates JetStream consumers and replays deletion markers even when no users need synchronization. Share one live key index per worker so initialized idle polls require no NATS requests. Replay initial snapshots in batches of 256 to prevent subscription overflow with large queues or slow readers.

- Resume ordered live updates after the snapshot's last revision, preserving changes made during the handoff.
- Allow long snapshots to finish while they make progress; rebuild interrupted snapshots and remove temporary consumers on completion or cancellation.
- Expose locally created claims immediately so expired work remains recoverable when watch notifications are delayed.
- Clear queues from cached candidates and newer stream revisions, avoiding deletion-history replay when nodes reconnect while including writes whose notifications are delayed.
- Bound bulk queue operations to 32 concurrent writes per store.
- Compact completed-claim deletion markers on the scheduler leader, preserving pending work and concurrently recreated keys.
- Restore application logger states after test migrations so log assertions remain independent of collection order.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I checked database migrations when models or schema changed. No schema changes.

## Testing

With `DEBUG=false`, `TZ=UTC`, and `NATS_SERVER_BINARY` pointing to a local NATS server:

```sh
python -m pytest tests/test_nats_kv_index.py tests/test_nats_node_memory.py tests/test_nats_sync_integration.py -q
```

- NATS suite: 35 passing tests against NATS 2.10.29, including four independent worker processes, claim expiry, reconnects, server restart, consumer recreation, bounded snapshot delivery, cancellation, and snapshot-to-live handoff races.
- Full local SQLite suite: 631 passed, zero skipped, with the external network tests enabled.
- Local stress run: 3.5 million deletion markers replayed by four readers with no dropped messages, four steady consumers, and zero outgoing NATS messages from initialized idle polls. Twenty subsequent clears received only 60 control responses without replaying the old markers.
- Ruff lint, formatting checks, and `git diff --check` passed for the changed files.

## Screenshots

Not applicable.

## Notes for reviewers

The index discovers candidate keys; authoritative values and claim ownership still use JetStream reads and CAS operations. Acknowledged local writes are immediately discoverable while watch notifications catch up. Compaction only removes revisions through an observed deletion marker after a five-minute grace period. The watcher is closed during bridge shutdown.

The bounded snapshot adapter uses the KV object's underlying JetStream context because the KV watch API exposes only push delivery. Real-server tests cover the snapshot consumer lifecycle and ordered live subscription recovery.

## Summary by CodeRabbit

- **New Features**
  - Improved node synchronization with faster live queue tracking.
  - Added automatic cleanup of stale synchronization records.
  - Added recovery for interrupted synchronization and service restarts.

- **Performance & Reliability**
  - Improved handling of concurrent user updates with controlled parallelism.
  - Reduced unnecessary queue polling and resource usage.
  - Preserved pending updates during cleanup and reconnection scenarios.

- **Tests**
  - Added extensive coverage for synchronization, failover, cleanup, concurrency, and restart behavior.


x0sina:
@coderabbitai review

DrSaeedHub:
Superseded by #892, which consolidates the complete user-sync performance and reliability fixes with regression tests into one commit targeting pref-boost.

PR #896, opened 2026-09-12
fix(node): make user synchronization reliable and avoid redundant work

DrSaeedHub:
## Summary

New or edited users can remain absent or stale on nodes when a queue wake-up is lost, a full snapshot overlaps pending updates, or database commits and queue writes arrive in different orders. Repeated edits with unchanged node settings also generate unnecessary work for every node.

- Make pending updates durable and revision guarded: preserve active claims, acknowledge only the claimed revision, and recover interrupted or expired work without resurrecting stale payloads.
- Read current user state immediately before delivery. Coalesce reads within each worker process, use indexed batches of up to 400 user IDs, and bound database and delivery timeouts.
- Coordinate full snapshots with a renewable per-node fence. Wait for in-flight deltas, load the snapshot inside the fence, and retire captured revisions only after a successful sync.
- Skip node dispatch when the serialized credentials and inbound set are unchanged. Preserve normal database writes, edit timestamps, notifications and API responses; changes to the effective node state still propagate.
- Avoid loading full user snapshots during healthy reconnect recovery. Resume shared work, bound queue batches, retain chunked delta transport, and compact expired KV history with revision-limited batched purges.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [x] Tests / CI

## Checklist

- [x] I updated documentation, translations, or examples if needed. (Operational considerations are described below; no UI changes.)
- [x] I checked database migrations when models or schema changed. (No schema changes; migration upgrade and consistency checks passed.)

## Testing

Validated head: `b6f8846f9cb3e9efa3fa0730b03c4452d8140ff8`.

| Validation | Result |
| --- | --- |
| Full suite with SQLite | 743 passed, zero failures/errors/skips |
| Full suite with PostgreSQL 17, four configured workers and real NATS JetStream | 743 passed, zero failures/errors/skips |
| Focused synchronization and recovery suites | 108 passed |
| Real NATS, gRPC transport and independent-process integration suites | 32 passed |
| Alembic upgrade and consistency checks | Passed |
| GitHub Actions on this head: SQLite, PostgreSQL, TimescaleDB, MySQL, MariaDB, Ruff and Python CodeQL | Passed |
| Ruff lint/format checks on all 23 changed Python files; `git diff --check` | Passed |

Full-suite command: `python -m pytest tests -q -rs --tb=short -p no:cacheprovider -o faulthandler_timeout=300`, with isolated databases, `TZ=UTC`, `DEBUG=false`, and the appropriate NATS/worker configuration. Migration checks used `python -m alembic upgrade head` and `python -m alembic check` against the isolated databases.

Regression coverage includes lost wake-ups and idle-worker races; cross-process claims, expiry and crash recovery; concurrent snapshots and updates; failed RPCs and fence renewal; reversed commit/enqueue order; bounded current-state queries and cancellation; unchanged versus effective user edits; healthy reconnect recovery; and concurrent key recreation during compaction.

Controlled deployment validation of the same runtime files:

- On 12 enabled nodes, create, credential rotation, disable, re-enable and delete were verified through actual WebSocket and Reality connections. The lifecycle checks also passed during a bulk edit workload; API-start-to-confirmation bounds were below five seconds, including client setup and probing overhead.
- Twelve repeated same-group edits produced zero owned queue writes, compared with 296 before suppression. A note-only edit dropped from 24 writes to zero; a credential change still reached all 12 nodes.
- Separate continuity checks kept 12 existing connections open across edits to another user: 394/394 requests succeeded.
- A 35-minute observation recorded all 12 nodes connected in 70 samples, no sync errors or core restarts, and no pending work in 18 detailed queue samples. A bulk workload included 18,182 successful user edits; before the concurrent lifecycle probe began, 15,510 modification events produced no sync-stream writes.

The deployment checks cover the 12 enabled nodes and the specified transports and observation window. The GitHub checks listed above also passed on this head.

## Screenshots

Not applicable.

## Notes for reviewers

Public API contracts, dependencies, deployment configuration and database schema are unchanged. Full administrative syncs retain their existing node-core restart behavior; ordinary user updates use the delta path. Automatic healthy recovery preserves queued work and avoids restarting the core.

Shared KV documents now carry claim metadata and refresh markers. Legacy claim records are recovered after lease expiry. Upgrade workers together; for a downgrade, stop workers and resolve refresh markers with the newer implementation before starting older workers. This change provides retryable delivery and convergence, not exactly-once transport.

### DIFF
--- a/app/node/bridge.py
+++ b/app/node/bridge.py
@@ -1,0 +1,47 @@
+"""Keep bounded background sync batches on the node's delta-update endpoint."""
+
+from contextvars import ContextVar
+
+from PasarGuardNodeBridge import NodeType, PasarGuardNode
+from PasarGuardNodeBridge.common.service_pb2 import User
+from PasarGuardNodeBridge.grpclib import Node as GrpcNode
+from PasarGuardNodeBridge.rest import Node as RestNode
+
+_queued_sync: ContextVar[bool] = ContextVar("queued_node_user_sync", default=False)
+
+
+class _QueuedBatchSync:
+    async def _sync_worker(self):
+        token = _queued_sync.set(True)
+        try:
+            await super()._sync_worker()
+        finally:
+            _queued_sync.reset(token)
+
+    async def _sync_batch_users(self, users: list[User]) -> list[User]:
+        # The bridge normally switches to per-user RPCs below 1000 users.
+        # Bounded queue claims still need the efficient delta-batch endpoint.
+        # Keep explicit per-user fallbacks outside the background worker intact.
+        if _queued_sync.get() and users:
+            supported, _ = await self._supports_chunked_sync()
+            if supported:
+                return await self.sync_users_chunked(
+                    users, chunk_size=min(100, len(users)), flush_pending=False, timeout=self._internal_timeout
+                )
+        return await super()._sync_batch_users(users)
+
+
+class QueuedGrpcNode(_QueuedBatchSync, GrpcNode):
+    pass
+
+
+class QueuedRestNode(_QueuedBatchSync, RestNode):
+    pass
+
+
+def create_node(connection: NodeType, **kwargs) -> PasarGuardNode:
+    if connection is NodeType.grpc:
+        return QueuedGrpcNode(**kwargs)
+    if connection is NodeType.rest:
+        return QueuedRestNode(**kwargs)
+    raise ValueError("invalid backend type")
--- a/app/jobs/node_checker.py
+++ b/app/jobs/node_checker.py
@@ -1,4 +1,5 @@
 import asyncio
+import time
 
 from PasarGuardNodeBridge import Health, NodeAPIError, PasarGuardNode
 from PasarGuardNodeBridge.storage import LifecycleStatus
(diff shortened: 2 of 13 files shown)


ImMohammad20000:
@coderabbitai review

app/node/bridge.py
```diff
@@ -0,0 +1,47 @@
...
+from PasarGuardNodeBridge.rest import Node as RestNode
+
+_queued_sync: ContextVar[bool] = ContextVar("queued_node_user_sync", default=False)
+
+
+class _QueuedBatchSync:
+    async def _sync_worker(self):
+        token = _queued_sync.set(True)
+        try:
+            await super()._sync_worker()
```
ImMohammad20000: the class not inherited anything how you use super class

ImMohammad20000:
pr the bridge part to https://github.com/PasarGuard/node_bridge_py

app/node/bridge.py
DrSaeedHub: `_QueuedBatchSync` is used as a mixin here. The concrete classes inherit from both `_QueuedBatchSync` and `GrpcNode`/`RestNode`:

```python
class QueuedGrpcNode(_QueuedBatchSync, GrpcNode):
    pass

class QueuedRestNode(_QueuedBatchSync, RestNode):
    pass
```

So `super()` follows the MRO and resolves `_sync_worker()` from `GrpcNode` or `RestNode`. `_QueuedBatchSync` is not intended to be instantiated on its own.

DrSaeedHub:
> pr the bridge part to https://github.com/PasarGuard/node_bridge_py
i did
https://github.com/PasarGuard/node_bridge_py/pull/21

x0sina:
@coderabbitai review

PR #903, opened 2026-09-15
fix(node): let lifecycle operations preempt user sync

dr-hoseyn:
## Summary

Node edits, disables, and removals could wait for many minutes when a bulk or full user sync held the per-node lock. A failed chunked sync made this especially visible because the bridge fallback retried users individually while lifecycle requests remained queued.

- Track bulk and full user-sync tasks per node.
- Let real configuration changes and removals cancel stale sync work before taking the lifecycle lock.
- Preserve unchanged health-check reconnects without interrupting an active sync.
- Release the sync lock between bounded batches and resolve the current managed node for every batch, preventing work from continuing on a replaced instance.
- Keep each per-node lock identity stable for the manager lifetime so concurrent waiters cannot bypass one another through a replacement lock.

The PR changes only `app/node/__init__.py`; it adds no documentation, test files, migrations, dependencies, or unrelated changes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation
- [ ] Tests / CI

## Checklist

- [ ] I added or updated tests for behavior changes. (No test files requested for this hotfix; existing suites plus isolated concurrency regressions were run.)
- [x] I updated documentation, translations, or examples if needed. (Not needed; no public behavior or configuration change.)
- [x] I checked database migrations when models or schema changed. (No model or schema changes.)

## Testing

- `python -m pytest tests/test_node_manager.py tests/test_node_manager_sync.py -q` — 11 passed.
- `python -m ruff check --no-fix app/node/__init__.py` — passed.
- `python -m ruff format --check app/node/__init__.py` — passed.
- `git diff --check` — passed.
- Isolated async regression: a legacy per-user fallback blocked indefinitely, then concurrent removal cancelled it and completed in 0.0001s.
- Isolated async regression: an unchanged reconnect preserved the in-flight sync, while a real config update cancelled it, replaced the node, and completed in 0.0003s.

## Notes for reviewers

Cancellation is limited to user-sync tasks for the affected node and only happens for an actual configuration change or lifecycle removal. Automated reconnects with an unchanged connection signature and metadata continue to reuse the live node without disrupting synchronization.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node lifecycle handling during user synchronization.
  * In-progress synchronization tasks are now canceled when node configurations change or nodes are removed.
  * User synchronization now rechecks node availability between batches, preventing updates to unavailable or outdated nodes.
  * Improved reliability when configuration changes occur during synchronization.


dr-hoseyn:
@coderabbitai review

PR #905, opened 2026-09-16
fix(core-editor): allow loopback tunnel targets

### LINKED ISSUE #895: [Bug]: protocol tunnel target address of 127.0.0.1 is not allowed (?)
### Before you submit

### PasarGuard version

v5.4.0

---

### Installation type

Docker Compose

---

### What happened?

if we add a tunnel protocol , the target address of `127.0.0.1` is cleared out when added to its input field,
if we leave it empty, it cannot be saved, shows error "Target address is required", 

---

### Steps to reproduce

1. add inbound type tunnel
2. add target address of `127.0.0.1` or leave it empty (try to save it)

---

### Expected behavior

xray-core accepts `127.0.0.1` as a target address

if `127.0.0.1` is the default, it should not complain about not being set, if `127.0.0.1` is not the default address, the `127.0.0.1` should be allowed to set

---

### Logs and error messages

```shell
there was not any log
```

---

### Environment

docker compose , on linux

### Relevant configuration

```yaml

```

### Screenshots or videos

_No response_

---

### Additional context

adding 127.0.0.1 is cleared out in input field , here I added 127.0.0.1 and it disappears  

Image

leaving it to empty, trying to save (assuming 127.0.0.1 will be the default) shows error 

Image

dr-hoseyn:
## Summary
- preserve explicit loopback addresses in tunnel and dokodemo-door inbounds
- allow an empty target address so Xray can apply its localhost default
- keep loopback values visible while editing instead of clearing the field

## Validation
- `npm run build`
- verified empty, `127.0.0.1`, `localhost`, and `::1` through the form schema and strict Xray config compilation

Fixes #895

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved valid tunnel rewrite addresses, including loopback and legacy address values, when loading or migrating configurations.
  - Tunnel configuration forms now retain entered rewrite addresses instead of clearing recognized placeholder values.
  - Updated tunnel validation to continue requiring a destination port while allowing the rewrite address field to remain empty.


dr-hoseyn:
@coderabbitai review

PR #907, opened 2026-09-17
fix(dashboard): copy links from menus on insecure contexts

### LINKED ISSUE #890: [Bug]: Links won't copy to the clipboard if the connection is insecure
### Before you submit

### PasarGuard version

v5.3.0

### Installation type

Other

### What happened?

When clicking the 'Copy links' button, the base64 links are present in the server response (as seen in the browser's Network tab), but they are not being copied to the clipboard.

### Steps to reproduce

1. Set the variables to use the panel via a local IP or socket:
`UVICORN_HOST= "10.88.0.3"`
or
`UVICORN_UDS="/var/lib/pasarguard/pasarguard.socket"`
2. Navigate to the Users section.
3. Click "Copy configuration" and choose "links" or "links (base64)".

### Expected behavior

After clicking the button, the links are copied to the clipboard.

### Logs and error messages

```shell
Upon clicking the button, I receive a notification that the links have been copied.
```

### Environment

OS: Debian 13
Database: PostgreSQL 17
Running via Podman.
Panel is hosted locally, no domain, using a local ip.

### Relevant configuration

```yaml

```

### Screenshots or videos

_No response_

### Additional context

I'm having trouble copying the links. They show up in the API response after the button click, but the clipboard is empty. Could it be an issue with navigator.clipboard? Subscription and file download links are working, so it seems to be specific to these links

Lingikaushikreddy:
Fixes #890.

## Cause

When the panel is served over plain HTTP (a LAN IP via `UVICORN_HOST`, or `UVICORN_UDS` behind a proxy), `window.isSecureContext` is false, so `useClipboard` skips `navigator.clipboard` and falls back to `document.execCommand('copy')` on a hidden textarea.

"Copy links" / "links (base64)" run that fallback from inside the users action `DropdownMenu`, while its Radix focus scope is still trapping focus. The trap pulls focus back from the textarea, so `select()` selects nothing. `execCommand('copy')` still returns `true`, so the success toast shows but the clipboard is unchanged. That matches the report. The subscription URL `CopyButton` works because it isn't inside the menu.

## Fix

- `useClipboard`: in the fallback, first set the text from a `copy` event (`clipboardData.setData` + `preventDefault`) and call `execCommand('copy')`. This doesn't depend on focus or selection. The textarea path stays as a last resort, and the secure-context `navigator.clipboard` path is unchanged.
- "Copy user ID" in the same menu called `navigator.clipboard.writeText` directly, which throws on insecure contexts. It now goes through the same hook.

## How I verified

A reduced repro in headless Chrome (Playwright, `channel: 'chrome'`), with the page loaded from `http://192.168.x.x` so `isSecureContext === false`, and a `focusin` handler that returns focus to the menu like Radix `FocusScope` does. The clipboard was read back from a `localhost` page:

| | with focus trap | without focus trap |
|---|---|---|
| current textarea fallback | returns `true`, clipboard **unchanged** | copies |
| copy-event fallback (this PR) | copies | copies |

Also on this branch: `VITE_BASE_API=/ bun run build` succeeds. `tsc -p tsconfig.app.json` reports the same number of errors as `dev`. ESLint on the two changed files goes from 11 to 10 problems (the removed one is an unused `error` binding).

Not changed: `api-key-modal.tsx` also calls `navigator.clipboard.writeText` directly, so it has the same insecure-context issue. I can move it to the hook here or in a follow-up.


Lingikaushikreddy:
To be precise about what I measured: the headless-Chrome run used a reduced page that imitates the focus trap, not the dashboard itself, so I want to back the claim about the real menu with something better than a resemblance.

Checked in the installed dependency (`@radix-ui/react-dropdown-menu` 2.1.24, what `^2.1.24` resolves to here): `modal` defaults to `true`, the menu passes that through as `trapFocus`, and `FocusScope` with `trapped` installs the `focusin` handler that pulls focus back into the menu. That is the behaviour the reduced page reproduced, and `action-buttons.tsx` renders `DropdownMenu` without overriding `modal`.

Happy to record a run against the real panel if you'd prefer that over the reduced repro.

